PR #14, opened 2019-10-12
Tylerta temperature

Tylerta1:
1st Version of temperature.c & tests


BeVolt/App/src/Temperature.c
```diff
@@ -4,12 +4,26 @@
...
 
 #include "Temperature.h"
 #include "config.h"
 
-uint16_t ModuleTemperatures[NUM_TEMPERATURE_BOARDS][20];	// Each board may not have 20 sensors. Look at config.h for how many sensors
+int16_t ModuleTemperatures[NUM_TEMPERATURE_BOARDS][20];	// Each board may not have 20 sensors. Look at config.h for how many sensors
 																												// there are on each board. 20 is just max that the board can handle
+int16_t ModuleTempStatus[NUM_TEMPERATURE_BOARDS];
+
+typedef enum chargeState {
```
chaseblock: For style, the enum should be capitalized, and the elements should be all-caps to avoid confusion.
```suggestion
typedef enum ChargeState {
```

BeVolt/App/src/Temperature.c
```diff
@@ -4,12 +4,26 @@
...
+	discharging,
```
chaseblock:
```suggestion
	DISCHARGING,
```

BeVolt/App/src/Temperature.c
```diff
@@ -4,12 +4,26 @@
...
+	charging
```
chaseblock:
```suggestion
	CHARGING
```

BeVolt/App/src/Temperature.c
```diff
@@ -4,12 +4,26 @@
...
+} chargeState;
```
chaseblock: This is never used. It should either be deleted, or the variable `chargingState` should be changed to this type in order to put it to good use. 

Changing the variable's type and the associated function arguments would probably make the most sense, since there is a very limited number of states that the `chargingState` is restricted to.

```suggestion
} ChargeState;
```

BeVolt/App/src/main.c
```diff
@@ -269,21 +269,121 @@ int main(){
...
+void checkModuleTemperatureTest(void);
+void checkIndividualSensorTest(void);
+void checkDangerTest(void);
+void checkMasterTest(void);						// To implement later. Able to check everything with easeee
 
+extern int16_t ModuleTemperatures[NUM_TEMPERATURE_BOARDS][20];
 
 
 int main(){
 	UART3_Init(9600);
```
chaseblock: @SijWoo Does printf use UART1 or UART3?

SijWoo: For master v1.1, it'll use UART3. Once everyone is done testing on master v1.0 and we have the nucleo PCB, I'm going to fix up UART.c to only contain code for v1.1 and deprecate the v1.0 board

BeVolt/App/inc/Temperature.h
```diff
@@ -11,14 +11,29 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <stdbool.h>
-#include "LTC2983.h"
 #include "SPI.h"
+#include "Voltage.h"
```
SijWoo: Delete Voltage.h

PR #16, opened 2019-10-21
Feature/voltage

SijWoo:
Let me know if you see anything that needs to be changed or improved. Voltage measurements and detection are working.

### DIFF
--- a/BeVolt/App/inc/config.h
+++ b/BeVolt/App/inc/config.h
@@ -27,7 +27,7 @@ typedef ErrorStatus Status;
 
 //--------------------------------------------------------------------------------
 // MAX and MIN limits of lithium ion cells
-#define MIN_VOLTAGE_LIMIT				2.5		// Under voltage limit (Voltage)	(actual min: 2.5V)
+#define MIN_VOLTAGE_LIMIT				2.7		// Under voltage limit (Voltage)	(actual min: 2.5V)
 #define MAX_VOLTAGE_LIMIT				4.0		// Over voltage limit (Voltage)		(actual max: 4.2V)
 
 #define MAX_DISCHARGE_TEMPERATURE_LIMIT	73.00	// Max temperature limit (Celcius)	(actual max: 75C)
--- a/BeVolt/App/inc/config.h
+++ b/BeVolt/App/inc/config.h
@@ -16,12 +16,12 @@ typedef ErrorStatus Status;
 
 //--------------------------------------------------------------------------------
 // Basic Parameters of BPS layout
-#define NUM_VOLTAGE_BOARDS				1			// Number of Voltage Slave Boards
+#define NUM_VOLTAGE_BOARDS				2			// Number of Voltage Slave Boards
 #define NUM_TEMPERATURE_BOARDS			3			// Number of Temperature Slave Boards
 
 //--------------------------------------------------------------------------------
 // Battery Pack layout
-#define NUM_BATTERY_MODULES				12		// Number of battery modules
+#define NUM_BATTERY_MODULES				24		// Number of battery modules
 #define NUM_TEMPERATURE_SENSORS			62		// Number of temperature sensors
 #define NUM_TEMP_SENSORS_PER_MOD		2		// Number of temperature sensors per battery module
 
--- a/BeVolt/App/inc/config.h
+++ b/BeVolt/App/inc/config.h
@@ -8,7 +8,7 @@
 
 #include "stm32f4xx.h"
 
-typedef enum {SAFE = 0, DANGER = 1} SafetyStatus;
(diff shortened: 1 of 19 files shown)


BeVolt/App/src/Voltage.c
```diff
@@ -60,21 +60,29 @@ ErrorStatus Voltage_UpdateMeasurements(void){
...
  */
 SafetyStatus Voltage_IsSafe(void){
 	for(int32_t i = 0; i < NUM_BATTERY_MODULES; i++){
 		uint16_t voltage = Modules[i / 12].cells.c_codes[i % 12];
-		if(voltage > MAX_VOLTAGE_LIMIT || voltage < MIN_VOLTAGE_LIMIT){
-			return DANGER;
+		
+		// VOLTAGE_LIMITS are in floating point. The LTC6811 sends the voltage data
+		// as unsigned 16-bit fixed point integers with a resolution of 0.00001
+		if(voltage > MAX_VOLTAGE_LIMIT * 10000){
```
chaseblock: That's an odd (even) number of zeros, and I would feel better if they were given their own definition (maybe `#define LTC6811_SCALING_FACTOR 10000` or something) so that we can eliminate the likelihood of mistakes.

BeVolt/App/src/main.c
```diff
@@ -84,14 +85,22 @@ void preliminaryCheck(void){
...
 			// Toggle Current fault LED
 		}
 		
-		if(!Voltage_IsSafe()){
+		if(Voltage_IsSafe() == OVERVOLTAGE){
+			// Toggle Voltage fault LED
+			LED_On(OVOLT);
+		}
+		
+		if(Voltage_IsSafe() == UNDERVOLTAGE){
```
chaseblock: It would be (marginally) faster to call this function once and then switch.

BeVolt/App/inc/config.h
```diff
@@ -8,26 +8,26 @@
 
 #include "stm32f4xx.h"
 
-typedef enum {SAFE = 0, DANGER = 1} SafetyStatus;
+typedef enum {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE = 3} SafetyStatus;
```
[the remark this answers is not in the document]
SijWoo: I agree that having misc options encourage code that isn't specific. This type was originally meant to standardize true or false. I didn't want future new members to keep track of multiple enums, so I wanted all the `IsSafe` functions to return the same type. You're right that we could have other status messages, but I can't have multiple enums redefine `SAFE` so what do you think we should do? We can also just `typedef int SafetyStatus` and create macro #define for the error codes.

BeVolt/App/inc/config.h
```diff
@@ -8,26 +8,26 @@
...
 typedef ErrorStatus Status;
```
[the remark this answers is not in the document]
SijWoo: This was created to standardize true or false but before I decided on having a SafetyStatus type. Forgot to delete

BeVolt/App/inc/config.h
```diff
@@ -8,26 +8,26 @@
...
 
 // NUCLEO or custom
 #define NUCLEO		0		// Change 1 if using nucleo, 0 if not
 
 //--------------------------------------------------------------------------------
 // Basic Parameters of BPS layout
-#define NUM_VOLTAGE_BOARDS				1			// Number of Voltage Slave Boards
+#define NUM_VOLTAGE_BOARDS				2			// Number of Voltage Slave Boards
```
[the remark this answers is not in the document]
SijWoo: This was for testing. That's a good idea

BeVolt/App/src/Voltage.c
SijWoo: That's a good idea

BeVolt/App/src/main.c
SijWoo: Sure, that'll make the grouping of checks consistent as well

SijWoo: Can't have enums as the condition statement for switch, maybe `typedef int SafetyStatus` is the way to go

BeVolt/App/inc/config.h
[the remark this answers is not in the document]
chaseblock: We could definitely have everything use a unified enum type to return error/status messages. That's a better way of doing things than what I was thinking. Let's make a change in the future to migrate the other drivers to more specific messages then.

SijWoo:
Config.h will have to be reorganized to align with new voltage/temperature setup but that will be done in another branch

BeVolt/App/src/main.c
```diff
@@ -93,14 +93,20 @@ void faultCondition(void){
 			// Toggle Current fault LED
 		}
 		
-		if(Voltage_IsSafe() == OVERVOLTAGE){
+		if(!Voltage_IsSafe()){
```
chaseblock: Since we're switching on this anyways, this is probably unnecessary, but whatever.

PR #17, opened 2019-10-21
Adds Working watchdog timer.

SijWoo:
Watch dog timer tested and working.

### DIFF
--- a/BeVolt/App/src/main.c
+++ b/BeVolt/App/src/main.c
@@ -14,6 +14,7 @@
 #include "EEPROM.h"
 #include "WDTimer.h"
 #include "SoC.h"
+#include "LED.h"
 
 void initialize(void);
 void preliminaryCheck(void);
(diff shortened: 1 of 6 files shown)


BeVolt/App/src/main.c
```diff
@@ -110,7 +115,7 @@ void faultCondition(void){
 // E.g. If you want to run a LTC6811 test, change "#define CHANGE_THIS_TO_TEST_NAME" to the
 //		following:
 //		#define LTC6811_TEST
-#define UART_TEST
+#define WATCHDOG_TEST
```
chaseblock: As a general suggestion, I'm not sure if we ever plan on moving to a Make-based system, but we should look into something that would allow us to remove these definitions from the src. When compiling from the terminal, `-DWATCHDOG_TEST` would handle it, and this would be easy to incorporate into a makefile. I realize Keil probably complicates this, but it might be worth looking into so that this line isn't changed every time a commit is made. 

The `master` branch should probably default to the actual `main()`. We should slate this as something to look into in the future.

PR #18, opened 2019-10-26
Chase

chaseblock:
Problem
=======
The current drivers and a reliable ADC needed to be implemented.

Solution
========
This includes an implementation of a DMA-based ADC and the current app drivers. The current driver  has been tested using a power supply with a resistor in series and appeared to give the correct values (within acceptable error) when compared to the DC supply's own measurements of the current.

Notes
=====
There are some interesting things going on in the ADC driver that were the final things to be fixed. This had to do with priority settings. Everything else is pretty straight-forward I think.

### DIFF
--- a/BeVolt/App/src/Current.c
+++ b/BeVolt/App/src/Current.c
@@ -11,15 +11,15 @@
 #include "ADC.h"
 #include "config.h"
 
-uint32_t HighPrecisionCurrent;	// Amp measurement of hall effect sensor of high precision
-uint32_t LowPrecisionCurrent;	// Amp measurement of hall effect sensor of high precision
+int32_t HighPrecisionCurrent;	// Amp measurement of hall effect sensor of high precision
+int32_t LowPrecisionCurrent;	// Amp measurement of hall effect sensor of high precision
 
 typedef enum {
 	HIGH_PRECISION,
 	LOW_PRECISION
 } CurrentSensor;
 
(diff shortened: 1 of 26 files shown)


BeVolt/Periphs/src/ADC.c
```diff
@@ -1,57 +1,104 @@
 /** main.c
  * Program for UTSVT BeVolt's Battery Protection System ADC initialization/output
- * @authors Russell Sutton, Garrett Wong, Sijin Woo
+ * @authors Garrett Wong, Sijin Woo, Russell Sutton, Chase Block
  * @lastRevised 10/26/2019
  */
 
 #include <stdint.h>
 #include "stm32f4xx.h"
+#include "stm32f4xx_dma.h"
```
SijWoo: You don't have to include stm32f4xx_dma.h. stm32f4xx.h includes everything for you already

BeVolt/App/src/Current.c
```diff
@@ -42,21 +60,49 @@ SafetyStatus Current_IsSafe(void){
  * @return 1 if charge, 0 if discharge
  */
 int8_t Current_IsCharging(void){
-	return 0;
+	// TODO: Make sure that the current board is installed in such a way that negative => charging
+	return HighPrecisionCurrent < 1;
```
SijWoo: Is this supposed be `< 0;` ?

chaseblock: It definitely is.

PR #19, opened 2019-10-26
Voltagewiredisconnect

JimothyGreene:
Methods to check for disconnected wires in the voltage boards. One to get summary, one to return the SafetyStatus, one to return a hex string to represent location of disconnects. Also created test in main.c


BeVolt/App/inc/Voltage.h
```diff
@@ -35,6 +35,23 @@ SafetyStatus Voltage_IsSafe(void);
  * @return pointer to index of modules that are in danger
  */
 uint16_t *Voltage_GetModulesInDanger(void);
+ 
+/** Voltage_OpenWireSummary
+ * Runs the open wire method with print=true
```
chaseblock: Since we're trying to abstract this out a bit, it might be better to include a high-level description of what we can expect the function to actually accomplish.

BeVolt/App/src/main.c
```diff
@@ -20,7 +20,8 @@ void initialize(void);
 void preliminaryCheck(void);
 void faultCondition(void);
 
-int mainmain(){
+int mainmain(){}
+int real_main(){
```
chaseblock: What's the point of having `mainmain` and `real_main`?

BeVolt/Drivers/inc/LTC6811.h
```diff
@@ -154,8 +154,10 @@ void LTC6811_adol(
 /*!  Start an open wire Conversion
 */
 void LTC6811_adow(
-  uint8_t MD, //!< ADC Conversion Mode
-  uint8_t PUP //!< Controls if Discharge is permitted during conversion
+	uint8_t MD, //ADC Mode
+  uint8_t PUP, //Discharge Permit
+	uint8_t CH, //Channels
+	uint8_t DCP//Discharge Permit
```
chaseblock: What is the difference between the discharge permits?

BeVolt/Drivers/inc/LTC6811.h
```diff
@@ -357,8 +359,9 @@ int16_t LTC6811_run_adc_redundancy_st(uint8_t adc_mode,
                                       cell_asic ic[]);
 
 /*! Helper function that runs the datasheet open wire algorithm*/
-void LTC6811_run_openwire(uint8_t total_ic,
-                          cell_asic ic[]);
+uint32_t LTC6811_run_openwire_multi(uint8_t total_ic,
+                          cell_asic ic[],
+													bool print);
```
chaseblock: This is a low-level driver. I don't think that this should depend on the UART driver in order to perform its function. I think that the better option here would be to have this driver return some information that a higher-level function could parse and output.

This might be alright, but just seems kind of odd to me. @SijWoo what do you think?

BeVolt/Drivers/inc/LTC681x.h
```diff
@@ -485,8 +491,9 @@ int16_t LTC681x_run_adc_redundancy_st(uint8_t adc_mode,
                                       cell_asic ic[]);
 
 /*! Helper function that runs the datasheet open wire algorithm*/
-void LTC681x_run_openwire(uint8_t total_ic,
-                          cell_asic ic[]);
+uint32_t LTC681x_run_openwire_multi(uint8_t total_ic,
+                          cell_asic ic[],
+													bool print);
```
chaseblock: See the comment on `LTC6811.h`

BeVolt/Drivers/src/LTC681x.c
```diff
@@ -468,8 +470,9 @@ void LTC681x_adstatd(
 
 
 // Start an open wire Conversion
+/*
 void LTC681x_adow(
```
chaseblock: Dead code should be cleaned up.

BeVolt/Drivers/src/LTC681x.c
```diff
@@ -1287,6 +1309,7 @@ int16_t LTC681x_run_adc_redundancy_st(uint8_t adc_mode, uint8_t adc_reg, uint8_t
 }
 
 //Runs the datasheet algorithm for open wire
+/*
 void LTC681x_run_openwire(uint8_t total_ic, cell_asic ic[])
```
chaseblock: Dead code should be cleaned up.

BeVolt/Drivers/inc/LTC6811.h
JimothyGreene: It can be removed, but it was just easier to be able to test it efficiently

BeVolt/App/src/Voltage.c
```diff
@@ -68,9 +68,8 @@ SafetyStatus Voltage_IsSafe(void){
...
 			return DANGER;
 		}
-	}
-
-	return DANGER;
+	} 
+	return SAFE;
 }
 
 /** Voltage_GetModulesInDanger
```
chaseblock: Let's get this function to flag any disconnected modules as dangerous.

PR #20, opened 2019-11-02
Eeprom future

chaseblock:
The EEPROM is working. We can write and read messages to/from the EEPROM in quick succession.

### DIFF
--- a/BeVolt/Drivers/inc/EEPROM.h
+++ b/BeVolt/Drivers/inc/EEPROM.h
@@ -12,14 +12,6 @@
 
 #define EEPROM_ADDRESS (0xA0)
 
-//starting addresses for errors
-#define EEPROM_FAULT_CODE_ADDR 0x00000// TODO: this was originally 0x00000. May want to change back.
-#define EEPROM_TEMP_FAULT  		 0x00100
-#define EEPROM_VOLT_FAULT		   0x04000
-#define EEPROM_CURRENT_FAULT	 0x08000
-#define EEPROM_WWDG_FAULT		   0x08100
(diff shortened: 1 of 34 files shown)


BeVolt/Drivers/inc/EEPROM.h
```diff
@@ -10,29 +10,82 @@
...
+/** EEPROM_LogData
+ * This logs the given data at an appropriate spot in memory.
+ * This does not log an actual error, but allows more specific
+ * information about failed modules to be logged.
+ */
+void EEPROM_LogData(uint8_t logType, uint8_t data);
+
 /** EEPROM_Save
  * Save some information to the EEPROM
+ * logType is type of fault (fault code)
```
SijWoo: Add attribute tags for parameters and return statements so we can generate doxygen documentation (similar to javadocs) in the future.
`@param logType : type of fault` or something like that

BeVolt/Drivers/inc/EEPROM.h
```diff
@@ -10,29 +10,82 @@
...
+ * This is required for clearing data and setting up new
+ * EEPROMs. This cmd really should be reserved for specialized
+ * set-up programs and the cli.
+ */
+void EEPROM_Reset(void);
+
+/** EEPROM_LogData
+ * This logs the given data at an appropriate spot in memory.
+ * This does not log an actual error, but allows more specific
+ * information about failed modules to be logged.
```
SijWoo: Add attribute tags for parameters and return statements so we can generate doxygen documentation (similar to javadocs) in the future.
`@param logType : type of fault
@param data : stored into EEPROM`

PR #21, opened 2019-11-16
Added timeouts to the waits in I2C peripheral

JimothyGreene:
Not sure why the RTE file shows changes.

The I2C read/write methods now returns an ErrorStatus that corresponds to the timeout counter.


BeVolt/Periphs/src/I2C.c
```diff
@@ -48,10 +48,16 @@ uint8_t I2C3_Read(uint8_t deviceAddr, uint16_t regAddr){
 	return data;
 }
 
-void I2C3_ReadMultiple(uint8_t deviceAddr, uint16_t startAddr, uint8_t *rxData, uint32_t rxSize){
+ErrorStatus I2C3_ReadMultiple(uint8_t deviceAddr, uint16_t startAddr, uint8_t *rxData, uint32_t rxSize){
+	int timeout_threshold = 240000000;
```
chaseblock: This should probably be either `#defined` or declared `const`, especially since all the functions use the same. Some information about how long the delay (in the form of a comment, etc.) would also be useful.

BeVolt/Periphs/src/I2C.c
```diff
@@ -141,40 +225,88 @@ void I2C3_WriteMultiple(uint8_t deviceAddr, uint16_t startAddr, uint8_t *txData,
...
+	timeout_count = 0;
 	while(!I2C_CheckEvent(I2C3, I2C_EVENT_MASTER_TRANSMITTER_MODE_SELECTED)) {
 		if(I2C3->SR1 & 0x0400) {
 			I2C3->SR1 &= ~0x0400;
 			I2C_GenerateSTOP(I2C3, ENABLE);
 			goto writePoll;
 		}
+		// Assume running at 80 MHz
+		timeout_count++;
+		// Returns and breaks after 3 seconds (240 million iterations)
```
chaseblock: The comment is incorrect.

PR #25, opened 2019-11-16
Updated Contactor.c functionality

JimothyGreene:
- Pin 1 now configured to an Input pin
- Contactor_Flag() now returns a FunctionalState enum


BeVolt/App/inc/Contactor.h
```diff
@@ -21,6 +22,6 @@ void Contactor_Off(void);
 
  
 //Outputs: flag status (0 or 1)
```
SijWoo: Just needs docmentation

PR #39, opened 2020-02-13
Refactor the build system with the new Makefile.

chaseblock:
Overview
========
This commit extends upon the introduction of the Linux
Makefile to make the build-system much cleaner. This is
done by:

1. Moving the "real" `main()` in `main.c` to be located at
the bottom of the file. In addition, the systemLoop function
was introduced at the top of the file for basic organization.
2. The creation of the `solar-build` script, which negates the
need to have the definition for the test that we want to run
in the actual files. In this way, we can build multiple targets
without needing to modify files.
3. Updating the README.md with instructions on how to get the
new build system working on a bash shell (assuming a Debian/Ubuntu
installation).
4. Fixing up some minor things to get the thing to build right.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "BeVolt/stlink"]
+	path = BeVolt/stlink
+	url = https://github.com/texane/stlink.git
--- a/BeVolt/App/src/main.c
+++ b/BeVolt/App/src/main.c
@@ -22,20 +22,37 @@ void initialize(void);
 void preliminaryCheck(void);
 void faultCondition(void);
 
-int realmain(){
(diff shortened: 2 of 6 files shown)


chaseblock:
So although I've opened this pull request, I think it's important to note that this __*WILL*__ screw up the test case build system in Keil by removing the line that we've been using to select the test case. So I think it's important that before we merge this, we make sure that everyone is familiar with this build system and we're sure we want to use it (although it's not hard to change back, it would be annoying).

That being said, I think that it would be good to hear what everyone has to think about this, even if we're not considering an immediate merge into master.

ClarkPoon:
Closing until the build system is complete.

PR #41, opened 2020-02-29
fixed Current_CheckStatus

AleNeff:
Current_CheckStatus checks that LowPrecisionCurrent is within safe limits based on whether or not override is set - now takes boolean override as input -- test added to main.c to print output of SAFE or DANGER based on LowPrecisionCurrent value & override parameter

### DIFF
--- a/BeVolt/App/src/Current.c
+++ b/BeVolt/App/src/Current.c
@@ -41,17 +41,16 @@ ErrorStatus Current_UpdateMeasurements(void){
  * Checks if pack does not have a short circuit
  * @return SAFE or DANGER
  */
-SafetyStatus Current_CheckStatus(void){
+SafetyStatus Current_CheckStatus(bool override){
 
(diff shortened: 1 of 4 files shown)


BeVolt/App/src/Current.c
```diff
@@ -41,25 +41,24 @@ ErrorStatus Current_UpdateMeasurements(void){
...
  */
-SafetyStatus Current_CheckStatus(void){
+SafetyStatus Current_CheckStatus(bool override){
 
-	if(HighPrecisionCurrent > MAX_HIGH_PRECISION_CURRENT) {
-		return (LowPrecisionCurrent < MAX_CURRENT_LIMIT)
-			? SAFE
-			: DANGER;
-	} else {
+	//if((LowPrecisionCurrent < MAX_CURRENT_LIMIT)&&(!override))
```
JimothyGreene: Unnecessary dead code

PR #43, opened 2020-03-06
Fixed probable cause of false current display from the CLI

JimothyGreene:
The CLI Current was displaying 1/1000 of the return from the Current library. I believe when I first wrote that, the Current library returned mA values. Since the CLI was only displaying to 2 decimal places, it was only showing 0.00A. This should fix it.


chaseblock:
It appears that the current library is still returning mA values. There seems to be some inconsistency with the units in the comments of that file. I would test before merging this.

chaseblock:
Upon further testing, this seems unnecessary.

PR #44, opened 2020-03-07
Coordinate a BPS-to-Dashboard safety message.

sugamarora99:
Enabled TIM2 Timer interupts to send a "danger" or "safe" message every half second to dashboard depending on override status. This will be used to communicate to the Dashboard information regarding the charging mode so that the motor can be disabled when needed.


BeVolt/App/src/main.c
```diff
@@ -14,8 +14,10 @@
 #include "SoC.h"
 #include "LED.h"
 #include "SysTick.h"
+#include "TIM2.h"
```
chaseblock: This no longer exists after the file rename, so needs to be changed.
```suggestion
#include "StatusSync.h"
```

BeVolt/App/src/main.c
```diff
@@ -231,7 +234,7 @@ void faultCondition(void){
 // E.g. If you want to run a LTC6811 test, change "#define CHANGE_THIS_TO_TEST_NAME" to the
 //		following:
 //		#define LTC6811_TEST
-#define CLI_TEST
+#define DashboardMessage
```
chaseblock: Definitions are best used in all uppercase to easily identify. I suggest renaming to something like this.
```suggestion
#define DASHBOARD_MSG
```

BeVolt/App/src/main.c
```diff
@@ -1344,4 +1347,32 @@ int main(void){
 	while(1){}
 }
 
+#elif defined DashboardMessage
```
chaseblock: See the above comment.
```suggestion
#elif defined DASHBOARD_MSG
```

BeVolt/Periphs/inc/CAN.h
```diff
@@ -23,6 +23,8 @@
 #define CAN_ID_SOC_DATA								(CAN_ID_BPS | 0x06)
 #define CAN_ID_WDOG_TRIGGERED					(CAN_ID_BPS | 0x07)
 #define CAN_ID_ERROR									(CAN_ID_BPS | 0x08)
+#define CAN_ID_MOTORON								(CAN_ID_BPS | 0x09)
+#define CAN_ID_MOTOROFF								(CAN_ID_BPS | 0x0A)
```
chaseblock: @ClarkPoon, @dimembermatt What's the difference between these IDs? The code in this commit is only using one.

BeVolt/App/src/main.c
```diff
@@ -1344,4 +1347,32 @@ int main(void){
...
+
+void DashboardMessageTest(){
```
chaseblock: This looks pretty reasonable. This doesn't do much for us once we've merged it into the master branch though. It would probably be good to actually incorporate this into our real main function in some way (maybe as a function that can be called from `main()`. What you have here would probably be good enough if you moved it out of the `#ifdef` and into say `StatusSync.c` to be called from main.

BeVolt/Periphs/inc/CAN.h
ClarkPoon: Should only be one CAN ID (0x10A) called Motor Disable (per the CAN BUS IDs spreadsheet on the drive)

BeVolt/App/src/main.c
```diff
@@ -16,6 +18,7 @@
 #include "SysTick.h"
 #include "PLL.h"
 #include "CLI.h"
+#include "CAN.h"
 
 cell_asic Minions[NUM_MINIONS];
 bool override = false;		// This will be changed by user via CLI
```
ClarkPoon, line 24: Can you add a call to DashMessage() inside of the while loop in realmain() so it is called when the BPS is running during the race?

BeVolt/App/src/main.c
```diff
@@ -93,6 +96,7 @@ void initialize(void){
 	Voltage_Init(Minions);
 	Temperature_Init(Minions);
 	CLI_Init(Minions);
+	CAN1_Init(CAN_Mode_Normal);
```
ClarkPoon, line 100: Can you add your initialization function to initialize() so that everything your functions need will be initialized when the BPS is run during the race?

chaseblock:
What's the holdup on this?

sugamarora99:
> @sugamarora99 the CAN messages were sending properly every 500 ms when we tested on the actual hardware in March, right?

No, when I tested this in March with the hardware it was sending them every .5 ms, which is why I made the counter go to 1000. After that week corona happened so this hasn't been tested yet.

ClarkPoon:
> > @sugamarora99 the CAN messages were sending properly every 500 ms when we tested on the actual hardware in March, right?
> 
> No, when I tested this in March with the hardware it was sending them every .5 ms, which is why I made the counter go to 1000. After that week corona happened so this hasn't been tested yet.

Ok, I will wait until we can test this before we merge it

ClarkPoon:
This has been solved by #260

PR #45, opened 2020-04-26
V2 of BPS code designed for portability with any chip/computer

SijWoo:
EVERYTHING COMPILES :))))

**Major modification:** BSP src files (Board support/specific package)
All hardware independent and dependent code was separated. Hardware dependent code includes chip specific I/Os, pins, architecture. Hardware independent code includes apps, drivers, and any code that calls the BSP src files. All hardware dependent code is identifiable with the BSP_ prefix in the file name. This separation will make code integration for new chips easier since the only files that'll have to be edited will be BSP_<src>.c files.

**Code change between v1.0 and v2.0** 
I created a tag for v1.0 in case we need to go back to it. I'm calling this new BPS code setup "v2". 

I didn't edit any code except for the UART interrupt stuff. All other code like SPI, I2C, etc. were only copied and pasted to their respective BSP definitions. I combined Fifo and UART code into one src file because it may be difficult for people when they add support for a different chip. I tested the new UART interrupts for STM32F413 on my own board so whatever worked in v1.0 will work in v2.0.

**What should we do with the opened branched?**
Idk if there are any good solutions to this. I think we should just go ahead with this change to begin the simulator. I don't see any major feature branches except for the balancing. We should let the people working on stuff in different branches know to move their code to the new branch.

**What's next?**
Our main.c has that test section that's absolutely horrendous. I made that with good intentions originally but i was too naive x.X I say for now on, we create different test files per module. I'll set up an easy way to add test cases with our new file format.


BeVolt/Apps/Src/CLI.c
```diff
@@ -539,14 +557,14 @@ void CLI_Watchdog(int* hashTokens) {
...
-	if(hashTokens[1] == NULL) {
+	if((int *)hashTokens[1] == NULL) {
 		EEPROM_SerialPrintData();
 		return;
 	}
 	uint8_t errorAddrArray[2];	// Need to get address where the error is stored
 	uint16_t errorAddr = 0;
 	switch(hashTokens[1]) {
-		case RESET:
+		case 0:
```
JimothyGreene, line 567: this case should be CLI_RESET_HASH

PR #66, opened 2020-05-28
I2 c simulator

ClarkPoon:
Added support for EEPROM (via BSP_I2C.c) to the simulator. EEPROM is simulated as a text file where each address is separated by a '\n' character, so the EEPROM can be easily read using a text editor. The simulated EEPROM can also be written to and read from using the functions in EEPROM.c and BSP_I2C.c


SijWoo:
noice

PR #67, opened 2020-05-29
CAN Simulation

sugamarora99:


BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,78 @@
...
+ * @param   data : data to be transmitted. The max is 8 bytes.
+ * @param   length : num of bytes of data to be transmitted. This must be <= 8 bytes or else the rest of the message is dropped.
+ * @return  0 if module was unable to transmit the data onto the CAN bus. Any other value indicates data was transmitted.
+ */
+
+uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
+    uint32_t offset;
+    // TODO: Transmit the data onto the CAN bus with the specified ID.
+    
+
```
ClarkPoon, line 38: Can you format the file with the CAN messages in a more human-readable format? Maybe something like
CAN id: 0x???
data: 0x??, 0x??, etc.
You also don't need to put the length of the CAN message, but if you want to, you could do something like this:
length: (some number)

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,78 @@
...
+ * @brief   Gets the data that was received from the CAN bus.
+ * @note    Non-blocking statement
+ * @pre     The data parameter must be at least 8 bytes or hardfault may occur.
+ * @param   id : pointer to store id of the message that was received.
+ * @param   data : pointer to store data that was received. Must be 8bytes or bigger.
+ * @return  0 if nothing was received so ignore id and data that was received. Any other value indicates data was received and stored.
+ * clear the message that was read from the file
+ */
+uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
+
```
ClarkPoon, line 60: This function will probably have to be updated if you change the CAN_Write function

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,78 @@
...
+ */
+
+uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
+    uint32_t offset;
+    // TODO: Transmit the data onto the CAN bus with the specified ID.
+    
+
+    FILE* fp = fopen(file, "w");
+    fprintf(fp, "0x%.3x,", id);
+    for(int i = 0; i < 8; i++){
```
ClarkPoon, line 41: In this for loop, you can check (i < length) instead of (i < 8). It is probably best to check both of them, in case someone passes a length greater than 8, so you do not go out of bounds on the data array

BeVolt/BSP/Simulator/Src/BSP_CAN.c
JimothyGreene: Since these are CSV files it makes it easier to read and write if there aren’t any row names. I feel like since the python simulator will be reading and displaying the data we shouldn’t be viewing the CSV files anyway. Just my thoughts but up to you.

BeVolt/BSP/Simulator/Src/BSP_CAN.c
sugamarora99: It seems that .csv files do not have an EOF, so all 8 locations of the data[] parameter for CAN_Read() get filled even if < 8 bytes are printed in CAN_Write()

BeVolt/BSP/Simulator/Src/BSP_CAN.c
ClarkPoon: That's a good point. Maybe just separate all of the data bytes with commas?

ClarkPoon: We will need a method in the .py file to parse the csv file if we are going to use on python to read the values

BeVolt/BSP/Simulator/Src/BSP_CAN.c
ClarkPoon: It is fine if the extra locations of data[] get filled with garbage values (this is probably what happens on the microcontroller) since we can tell which locations are valid based on the CAN id. If trying to change this causes a seg fault when CAN_Read() is called (trying to read past end of file), I would just leave it like this.

BeVolt/BSP/Simulator/DataGeneration/CAN.py
```diff
@@ -0,0 +1,13 @@
+import csv
+
+#path name to file
+file = "Data/CAN.csv"
+
+"""
+@brief send a CAN message as a specific system
+    Function called by simulate.py
+"""
+def Send_Message(id, message[8], length):
```
JimothyGreene, line 10: Make sure that you add the proper functionality to the simulate.py engine so that you can actually test this. This code won't run properly due to a syntax error here. In python, ```message[8]``` isn't a valid parameter. Everything is dynamically typed in Python, so the parameter would just be ```message``` and it would be understood that that is supposed to be a list. 

Also, I think we ended up deciding that you actually didn't need to put the row labels in the file since it's a CSV and not viewed by us anyway.

BeVolt/BSP/Simulator/DataGeneration/Data/CAN.csv
```diff
@@ -0,0 +1 @@
+0x10a,0182153675000000,5
```
JimothyGreene, line 1: I'm not exactly sure how this file got into your commit? The gitignore has *.csv on it

BeVolt/Tests/Test_BSP_CAN.c
```diff
@@ -0,0 +1,14 @@
...
+ int main(void){
+   uint8_t message[8] = {0x01, 0x82,0x00,0x36,0x75,0,0,0x96};
+   uint8_t readData[8] = {0,0,0,0,0,0,0,0};
+   uint32_t id[1];
+  
+   uint8_t length = 8;
+   
+   BSP_CAN_Write(0x10A, message, length);
+   BSP_CAN_Read(id, readData);
+}
```
JimothyGreene, line 14: Good test, but I would also test it using the CLI. There is functionality there to read CAN messages and also to write a CAN message to a specific id. So I'd just use that as well to test and make sure the BSP code works properly.

BeVolt/BSP/Simulator/DataGeneration/CAN.py
```diff
@@ -1,13 +1,28 @@
...
+        csvwriter = csv.writer(csvfile)
+        csvwriter.writerow([  "0x" + id, "0x"+message[0],"0x"+message[1],"0x"+message[2],"0x"+message[3],"0x"+message[4],"0x"+message[5],"0x"+message[6],"0x"+message[7] ])
+
+def Get_CAN_Info():
+    with open(file, 'r') as csvfile:
+        csvreader = csv.reader(csvfile)
+        return next(csvreader)
+
+def Invalid_CAN_ID(id):
+    ret = True
```
JimothyGreene, line 24: Unused variable

BeVolt/BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -53,16 +56,26 @@ def configure():
     # Get configuration settings
     global state, mode
     print("Welcome to the BPS Simulator")
-    print("Would you like to simulate 'charging' or 'discharging'?")
+    print("Would you like to simulate 'charging', 'discharging', or 'CAN'?")
```
JimothyGreene, line 59: I would keep CAN out of this section since 'charging' and 'discharging' are states that need to be set. Instead, you could add the CAN option to a new menu option during the KeyboardInterrupt. That way, when the user enters CTRL+C during execution, they can then have the option to simulate a CAN message. We should be able to do so while also simulating a charging/discharging battery.

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,84 @@
...
+ */
+uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
+    if(flag){
+        FILE* fp = fopen(file, "r+");
+        fseek(fp, 2, SEEK_SET);
+        fscanf(fp, "%3x", id);
+
+        fseek(fp, 6 ,SEEK_SET);
+        for(int i = 0; i < 8; i++){
+            fscanf(fp, "%*c%*c%2x%*c", data);
```
ClarkPoon, line 71: This line is causing a compiler warning "/BSP/Simulator/Src/BSP_CAN.c:71:33: warning: format ‘%x’ expects argument of type ‘unsigned int *’, but argument 3 has type ‘uint8_t * {aka unsigned char *}’ [-Wformat=]"

BeVolt/BSP/Simulator/DataGeneration/Data/CAN.csv
```diff
@@ -0,0 +1 @@
+0x103,0x20,0x20,0x20,0x20,0x20,0x20,0x20,0x20
```
ClarkPoon, line 1: Can you remove this file? Ideally, this file will only be stored locally and never committed

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
-    return 0;
-}
+#include "BSP_CAN.h"
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/file.h>
+
+static const char* file = "BSP/Simulator/DataGeneration/Data/CAN.csv";
+
+uint32_t flag;
```
ClarkPoon, line 8: Can you five the flag a more descriptive name or add a comment that explains what it is for?

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
+
+/**
+ * @brief   Gets the data that was received from the CAN bus.
+ * @note    Non-blocking statement
+ * @pre     The data parameter must be at least 8 bytes or hardfault may occur.
+ * @param   id : pointer to store id of the message that was received.
+ * @param   data : pointer to store data that was received. Must be 8bytes or bigger.
+ * @return  0 if nothing was received so ignore id and data that was received. Any other value indicates data was received and stored.
+ * clear the message that was read from the file
+ */
```
ClarkPoon, line 79: Whenever I run with the main() in Test_BSP_CAN.c, The id returned is incorrect. Not sure if this is an issue with CAN_Read() or CAN_Write(). I tested this by printing the id and data values after the call to CAN_Read() in the test main()

BeVolt/BSP/Simulator/DataGeneration/CAN.py
```diff
@@ -0,0 +1,34 @@
...
+
+def Get_CAN_Info():
+    os.makedirs(os.path.dirname(file), exist_ok=True)
+    with open(file, 'r') as csvfile:
+        fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)
+        csvreader = csv.reader(csvfile)
+        fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
+        return next(csvreader)
+
+def Invalid_CAN_ID(id):
```
chaseblock, line 30: This whole function can be rewritten as one line (which also likely performs better because of how python is optimized).
```python
def Invalid_CAN_ID(id):
    return not id in valid_IDs
```

BeVolt/BSP/Simulator/DataGeneration/Data/CAN.csv
chaseblock: I agree.

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
-    //      If message was received, then store the id and data of the message into the pointers and return a value other than 0.
-
-    return 0;
-}
+#include "BSP_CAN.h"
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/file.h>
+
+static const char* file = "BSP/Simulator/DataGeneration/Data/CAN.csv";
```
chaseblock, line 6: My opinion on this has been stated already. There should be a change in the simulator branch that allows for something like `static const char* file = GET_CSV_PATH(CAN_CSV_FILE);`

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
+ * @param   None
+ * @return  None
+ */
+
+
+
+void BSP_CAN_Init(void) {
+    FILE* fp = fopen(file, "w");
+    if(!fp) {
+        perror("CAN.csv");
```
chaseblock, line 21: This should also be changed to reflect that change above.

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
+ * @param   id : Message of ID. Also indicates the priority of message. The lower the value, the higher the priority.
+ * @param   data : data to be transmitted. The max is 8 bytes.
+ * @param   length : num of bytes of data to be transmitted. This must be <= 8 bytes or else the rest of the message is dropped.
+ * @return  0 if module was unable to transmit the data onto the CAN bus. Any other value indicates data was transmitted.
+ */
+
+uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
+    FILE* fp = fopen(file, "w");
+    if(!fp) {
+        perror("CAN.csv");
```
chaseblock, line 44: Likewise.

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
+ * @pre     The data parameter must be at least 8 bytes or hardfault may occur.
+ * @param   id : pointer to store id of the message that was received.
+ * @param   data : pointer to store data that was received. Must be 8bytes or bigger.
+ * @return  0 if nothing was received so ignore id and data that was received. Any other value indicates data was received and stored.
+ * clear the message that was read from the file
+ */
+uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
+    FILE* fp = fopen(file, "r+");
+    if(!fp) {
+        perror("CAN.csv");
```
chaseblock, line 83: Likewise.

BeVolt/BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -1,41 +1,116 @@
...
+            fscanf(fp, "%*c%*c%2hhx%*c", data);
+            data++;
+        }
+        flock(fno, LOCK_UN);
+        fclose(fp);
+    
+        
+        fp = fopen(file, "w");
+        if(!fp) {
+            perror("CAN.csv");
```
chaseblock, line 103: Likewise.

BeVolt/BSP/Simulator/DataGeneration/CAN.py
```diff
@@ -0,0 +1,34 @@
+import csv
+import fcntl
+import os
+
+
+file = "BSP/Simulator/DataGeneration/Data/CAN.csv"
```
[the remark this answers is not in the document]
JimothyGreene: You can take a look at what I did @sugamarora99 in the Controls repository. Look at the simulator_switches branch. I separated out the file path into a common header file.

chaseblock:
I like how this is looking. I've tested on Linux, and it seems to work, but the CAN messages display a little oddly if the terminal is too small. How does this look on Windows?

For reference, [this](https://chaseblock.com/wp-content/uploads/2020/06/StrangeDisplay.png) is what I'm talking about. The bit on the right straightens out when the window is enlarged, but the bit on the left remains. Perhaps some sort of wrapping would be best? We can change the default size of the simulator fairly easily, but that won't help if we need to support arbitrary-length messages.

BeVolt/BSP/Simulator/DataGeneration/CAN.py
```diff
@@ -56,12 +56,13 @@ def Send_Message(id, message, length):
         fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
 
 def Get_CAN_Info():
```
chaseblock, line 58: Wait for long enough, and this function will raise a StopIteration exception. This crashes the simulator. I've hit it a few times, usually after under one minute of running the simulator with the BSP_CAN test. 

This exception is standard of iterators: they raise this when they are out of data. My guess would be that instead of checking whether the file is non-empty, you need to check whether there is any data that has not already been read.

BeVolt/BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -37,8 +39,11 @@ def display(battery=None):
     stdscr.addstr(4, 0, f"Current:")
     stdscr.addstr(4, 10, f"{adc_values[1]} A ")
     #Read CAN data
-    stdscr.addstr(12, 80, f"                                                    ") #clear previous data
-    stdscr.addstr(12, 80, f"CAN: {CAN.Get_CAN_Info()}")
+    CANdata = CAN.Get_CAN_Info()
+    text = ' '.join([elem for elem in CANdata]) #put elements of the list of CAN data bytes into a string
```
chaseblock, line 43: Why not just this?
```suggestion
    text = ' '.join(CANdata) #put elements of the list of CAN data bytes into a string 
```

BeVolt/BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -104,10 +110,16 @@ def main():
...
         configure()
     global stdscr
+    global CANbox
     stdscr = curses.initscr()
     curses.start_color()
     curses.noecho()
     curses.cbreak()
+    #box is for CAN messages
+    CANbox = curses.newwin(7, 21, 12, 78)
+    CANbox.immedok(True)
```
chaseblock, line 120: My understanding is that `immedok` forces the window to update at any change, rendering refresh unnecessary. Probably only need to stick with one of these.

BeVolt/BSP/Simulator/DataGeneration/CAN.py
```diff
@@ -56,12 +56,13 @@ def Send_Message(id, message, length):
...
     os.makedirs(os.path.dirname(file), exist_ok=True)
     with open(file, 'r') as csvfile:
         fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)
         if os.stat(file).st_size == 0:     #check if file is empty
             fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
-            return 'Nothing has been sent through CAN bus'
+            return nothingOnCAN
         else:
             csvreader = csv.reader(csvfile)
             fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
```
chaseblock, line 68: My guess about this problem is that when this file lock is released, the BSP removes file contents before the next line can execute. Make sure the read the file before unlocking, then unlock and return properly.

BeVolt/BSP/Simulator/DataGeneration/CAN.py
chaseblock: I believe that I resolved this problem (see ad09785610d70cc50cd1aa84a7f93169a908e368).

PR #72, opened 2020-06-02
Contactor simulator

manthanand:
I wasn't sure how to test the on function because their isn't a command for turning on the contactor. Both the off and on functions are the same syntax so in order to test the function, I just changed the value in the Contactor_Off function from 0 to 1 and made sure the simulator read it as off and on respectively.

### DIFF
--- a/BeVolt/BSP/Simulator/DataGeneration/Contactor.py
+++ b/BeVolt/BSP/Simulator/DataGeneration/Contactor.py
@@ -7,11 +7,6 @@ file = "Data/Contactor.csv"
 
 #Reads file and returns 1 or 0, 1 is on, 0 is off
 def Read_Contactor():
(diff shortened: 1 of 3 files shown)


BeVolt/BSP/Simulator/DataGeneration/Contactor.py
```diff
@@ -0,0 +1,12 @@
...
+#This module will read from the csv file to determine the status of the Contactor
+
+#file path to data
+file = "Data/Contactor.csv"
+
+#Reads file and returns 1 or 0, 1 is on, 0 is off
+def Read_Contactor():
+    with open(file, 'r') as csvfile:
+        csvreader = csv.reader(csvfile)
+        return csvreader
```
JimothyGreene, line 12: This returns a ```reader``` object, not a state

PR #73, opened 2020-06-03
Contactor simulator

manthanand:
Debugged python file by running it independently and printing value in CSV file. It worked when I ran it and changed the values.


BeVolt/BSP/Simulator/DataGeneration/Contactor.py
```diff
@@ -0,0 +1,16 @@
...
+
+#file path to data
+file = "Data/Contactor.csv"
+
+#Reads file and returns 1 or 0, 1 is on, 0 is off
+def Read_Contactor():
+    with open(file, 'r') as csvfile:
+        csvreader = csv.reader(csvfile)
+        return next(csvreader)
+        #print(next(csvreader)): used this to test function
```
SijWoo, line 13: Go to the issues ticket title **Simulator: Locking files during read and writes** and add the necessary code to safely access the files.

BeVolt/BSP/Simulator/Src/BSP_Contactor.c
```diff
@@ -9,8 +11,13 @@
...
 void BSP_Contactor_Init(void) {
-    // TODO: Initialize two pins that will be connected to the Contactor. One must be configued as an input
-    //      and the other must be configued as an output.
+    /* Hardware: Initialize two pins that will be connected to the Contactor. One must be configued as an input
+          and the other must be configued as an output.
+    Software: Create file that contains one unsigned integer, 0 means off, 1 means on
+    */
+   FILE* fp = fopen(file, "w+"); //if file doesn't exist, it is created
+   fprintf(fp, "%d", 0); //write 0 to file
+   fclose(fp); //close file
```
SijWoo, line 20: Add a lock to prevent other programs from accessing/modifying the file

BeVolt/BSP/Simulator/Src/BSP_Contactor.c
```diff
@@ -20,8 +27,12 @@ void BSP_Contactor_Init(void) {
  * @return  None
  */
 void BSP_Contactor_On(void) {
-    // TODO: Set the state to high for the output pin.
+    // Hardware: Set the state to high for the output pin.
     //      Use Positive Logic.
+    // Software: Set integer to 1
+    FILE* fp = fopen(file, "w"); //Open file to write
+    fprintf(fp, "%d", 1); //Write 1 to file
+    fclose(fp);
```
SijWoo, line 35: Add lock

BeVolt/Apps/Src/CLI.c
```diff
@@ -1,774 +1,780 @@
...
+			break;
+		// Temperature of specific module
+		case CLI_MODULE_HASH:
+			if (hashTokens[2] == 0 || hashTokens[2]-1 > NUM_BATTERY_MODULES || hashTokens[2]-1 < 0){
+				printf("Invalid module number\n\r");
+			}
+			else {
+				if(hashTokens[3] == 0) {//temperature of module
+					printf("Module number %d: %.3f C\n\r", hashTokens[2], Temperature_GetModuleTemperature(hashTokens[2]-1)/MILLI_UNIT_CONVERSION);
+				} else if(hashTokens[3]-1 == 0 || hashTokens[3]-1 == 1) {//temperature of specific sensor in module
```
chaseblock, line 245: Why are we subtracting one and then comparing to 0 and 1? Why not `hashTokens[3] == 1` and `hashTokens[3] == 2`? In fact, since I'm not sure what 1 and 2 would mean for a hash token, why not use some sort of enum (if these are status codes)?

I realize this is probably out of scope for the contactor simulator pull, but we should address this at some point.

@JimothyGreene what's going on here?

BeVolt/Apps/Src/main.c
```diff
@@ -72,6 +72,8 @@ int main(){
 			break;
 		}
 
+		// TODO: Implement heartbeat for RUN light at a visible frequency
```
chaseblock, line 75: Can we make this an issue if it's not already? @ClarkPoon

BeVolt/BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -13,91 +14,146 @@
...
+    for i, module in enumerate(module_values):
+        stdscr.addstr(i+2, 37, f"{i+1}")
+        stdscr.addstr(i+2, 40, f"| {'X' if module[0] else ' '} | {module[1]/10000:.4f}V | {module[2]/1000:.3f}°C | {module[3]/1000:.3f}°C |")
     # Read LED values
+    stdscr.addstr(0, 90, "LEDs")
+    stdscr.addstr(1, 80, "=======================")
     lights = Lights.read()
-    for i in range(0, 9):
-        stdscr.addstr(i+2, 40, lights_names[i])
+    curses.init_pair(1, curses.COLOR_RED, curses.COLOR_RED)
```
chaseblock, line 49: Is this for the LEDs? Is this part of this set of changes, or did this come from another branch? If it came from elsewhere, we should make sure to review that branch's changes before doing anything here.

`git blame` shows that @SijWoo changed this line. Is this part of a finished feature?

BeVolt/Apps/Src/CLI.c
JimothyGreene: That's just a sensor number (0 or 1). I was just trying to keep the 0-indexing on the code side and the 1-indexing on the user input side. I can see how this is confusing, so it can be changed.

chaseblock: Yeah I think that we could either change the user input side to 0-indexed or change it so that we have definitions/constants like `SENSOR_ID_1 = 0`. Both would make it easier to read.

BeVolt/Apps/Src/main.c
ClarkPoon: Good catch, I just created a new issue for this

chaseblock:
Github seems to be having an aneurysm or something with this pull request. I'm approving, but I'm going to do the merge manually, rather than with github's interface, so that I can be sure nothing terrible happens.

According to a `git diff`, the only files that have actually changed are:

```
.vscode/settings.json                     |  5 ++++-
 BSP/Simulator/DataGeneration/Contactor.py | 22 ++++++++++++++++++++++
 BSP/Simulator/Src/BSP_Contactor.c         | 49 ++++++++++++++++++++++++++++++++++++++++++-------
 BSP/Template/Src/BSP_Contactor.c          | 49 ++++++++++++++++++++++++++++++++++++++++++-------
 4 files changed, 110 insertions(+), 15 deletions(-)
```

This appears safe to me.

PR #93, opened 2020-06-21
Voltage Library cell_asic fix

manthanand:
Created a private array inside Voltage.c for other functions to use. Tested with the simulator and all voltages were the same.


BeVolt/Apps/Src/Voltage.c
```diff
@@ -4,12 +4,12 @@
  */
 
 #include "Voltage.h"
-#include "LTC6811.h"
+#include "LTC681x.h"
```
SijWoo, line 7: Is there something you needed from LTC681x.h? The LTC6811 src file is a wrapper for the LTC681x src file so the LTC681x functions should not be used.

manthanand: No, I added it but forgot to remove it.

BeVolt/Apps/Src/Voltage.c
```diff
@@ -4,12 +4,12 @@
...
 #include "config.h"
 #include <stdlib.h>
 
 static cell_asic *Minions;
-
+uint16_t VoltageVal[NUM_BATTERY_MODULES]; //Voltage values gathered
```
chaseblock, line 12: If we're treating this as private, it should be `static`.

PR #94, opened 2020-06-21
Watchdog Timer Simulation

sugamarora99:


BeVolt/BSP/Simulator/Src/BSP_WDTimer.c
```diff
@@ -1,34 +1,95 @@
...
+#include <stdlib.h>
+#include <sys/file.h>
+#include <sys/types.h>
+#include <unistd.h>
 
+
+static const char* file = "BSP/Simulator/DataGeneration/Data/WDTimer.csv";
+
+bool DidSystemReset;
+bool HasStarted;
```
chaseblock, line 12: Declare these as static and rename them like `didSystemReset`.

BeVolt/BSP/Simulator/Src/BSP_WDTimer.c
```diff
@@ -1,34 +1,95 @@
...
-    // TODO: Start the watchdog countodwn
+    CheckReset();
+    FILE* fp = fopen(file, "w");
+    if(!fp) {
+        perror("WDTimer.csv");
+        exit(EXIT_FAILURE);
+    }
+    int fno = fileno(fp);
+    flock(fno, LOCK_EX);
+    fprintf(fp, "%x\n%d", 0xCCCC, (int)getpid());
```
chaseblock, line 89: What does 0xCCCC mean? If it is chosen for a specific purpose, we should create a `const int` or a `#define` for it. If it's not chosen for some particular reason, then why not use an enum?

BeVolt/BSP/Simulator/Src/BSP_WDTimer.c
```diff
@@ -39,4 +100,22 @@ void BSP_WDTimer_Start(void) {
...
+    if(HasStarted == true){
+        FILE* fp = fopen(file, "r+");
+        if(!fp) {
+            perror("WDTimer.csv");
+            exit(EXIT_FAILURE);
+        }
+        int fno = fileno(fp);
+        flock(fno, LOCK_EX);
+        
+        fprintf(fp, "%x\n%d", 0xAAAA, 0);
```
chaseblock, line 113: Likewise.

BeVolt/Tests/Test_BSP_WDTimer.c
```diff
@@ -0,0 +1,52 @@
+#include "common.h"
+#include "config.h"
+#include "BSP_WDTimer.h"
+
+int main(){
+    uint32_t delay = 150000000;
```
chaseblock, line 6: What unit is this in. Could we name it something like `delayMilliseconds` or `delayMicroseconds`?

BeVolt/BSP/Simulator/DataGeneration/WDTimer.py
```diff
@@ -18,16 +21,16 @@ def Check_State():
             fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)  
             csvreader = csv.reader(csvfile)
             currentState = next(csvreader)
-            if ((currentState[0] == 'aaaa') and (duration > 0)):
+            if ((currentState[0] == reset) and (duration > 0)):
                 duration = 15 #reset countdown
                 csvfile.seek(0)
                 csvwriter = csv.writer(csvfile)
```
chaseblock, line 27: csvfile is opened with `'r+'`, which will fail if the file is not there. Probably should change this to `'w+'`. Placing the comment here since Github won't let me comment on the relevant line. Look at line 20.

PR #95, opened 2020-06-23
Simulator strobelight

AleNeff:
Filepaths were changed for debugging purposes, should work otherwise - I think my VScode system is just jank in some subtle way that I haven't figured out yet. The state of the strobelight is stored in the txt file the src creates as a character that is "1" for pulsing and "0" for off.

### DIFF
--- a/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
+++ b/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
@@ -1,0 +1,19 @@
+import csv
+import os
+#This module holds status of strobe (on/off)
+
+#path name to file
+file = "BSP/Simulator/DataGeneration/Data/Strobe.csv"
+
+#returns integer where 1 means strobe is on, 0 means strobe is off
+def read();
+    strobe = 0
+    os.makedirs(os.path.dirname(file), exist_ok=True)
+    with open(file, 'a') as csvfile:
+        pass   #creates file if it doesn't exist
+    with open(file 'r') as csvfile: #read file
+        csvreader = csv.reader(csvfile)
+        if len(strobe):
+            return int (strobe)
+        else:
+            return 0
--- a/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
+++ b/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
@@ -6,15 +6,17 @@ import os
 file = "BSP/Simulator/DataGeneration/Data/Strobe.csv"
 
 #returns integer where 1 means strobe is on, 0 means strobe is off
-def read();
-    strobe = 0
+def read():
+    strobe = []
     os.makedirs(os.path.dirname(file), exist_ok=True)
     with open(file, 'a') as csvfile:
         pass   #creates file if it doesn't exist
-    with open(file 'r') as csvfile: #read file
+    with open(file, 'r') as csvfile: #read file
         csvreader = csv.reader(csvfile)
-        if row in csvreader:
-            strobe = row
-        return int (strobe)
+        for row in csvreader:
+            strobe.append(row)
+        if len(strobe):
+            return int(strobe[0][0])
         else:
             return 0
+print(read())
--- a/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
+++ b/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
@@ -19,4 +19,3 @@ def read():
             return int(strobe[0][0])
         else:
             return 0
-print(read())
--- a/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
+++ b/BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
@@ -13,7 +13,8 @@ def read();
         pass   #creates file if it doesn't exist
     with open(file 'r') as csvfile: #read file
         csvreader = csv.reader(csvfile)
-        if len(strobe):
-            return int (strobe)
+        if row in csvreader:
+            strobe = row
+        return int (strobe)
         else:
             return 0
--- a/BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
+++ b/BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
@@ -1,4 +1,4 @@
-#include "../../Inc/BSP_Strobelight.h"
+#include "BSP_Strobelight.h"
 #include <stdint.h>
 #include <stdlib.h>
 #define STROBE_OFF 0
@@ -14,8 +14,11 @@ int StrobeStatus = 0;
  */
 void BSP_Strobe_Init(void) {
 	FILE* fp = fopen(file, "w+");
+    int fno = fileno(fp);
+    flock(fno, LOCK_EX);
 	fprintf(fp, "0");
 	fclose(fp);
+    flock(fno, LOCK_UN);
 }
 
 /**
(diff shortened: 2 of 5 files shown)


BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
```diff
@@ -0,0 +1,43 @@
+#include "../../Inc/BSP_Strobelight.h"
```
JimothyGreene, line 1: You should just be able to put "BSP_Strobelight.h" in here. If you get an error from VSCode but not from the Makefile when you compile, it should be fine. The Makefile should have that directory on the include path.

BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
```diff
@@ -0,0 +1,43 @@
...
+#include <stdint.h>
+#include <stdlib.h>
+#define OFF 0
```
JimothyGreene, line 4: I would use specific names here such as STROBELIGHT_OFF and STROBELIGHT_PULSING

BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
```diff
@@ -0,0 +1,43 @@
...
+ * @brief   Turns off Strobe
+ * @param   None
+ * @return  None
+ */
+void BSP_Strobe_Off(void) {
+    StrobeStatus = OFF;     //0 = off
+    FILE* fp = fopen(file, "w");
+    fprintf(fp, "%d", StrobeStatus);
+    fclose(fp);
+}
```
JimothyGreene, line 43: Each time you open the file, add a lock to the file. Check out some of the other simulation branches to see how that works.

BeVolt/Tests/Test_BSP_Strobelight.c
```diff
@@ -0,0 +1,22 @@
+#include "../BSP/Inc/BSP_Strobelight.h"
```
JimothyGreene, line 1: Same as above with the include path

BeVolt/Tests/Test_BSP_Strobelight.c
```diff
@@ -0,0 +1,22 @@
...
+#include <stdio.h>
+
+static const char* file = "BSP/Simulator/DataGeneration/Data/Strobe.csv";
+char a;
+int  main() {
+BSP_Strobe_Init();
+BSP_Strobe_On();
+FILE* fp = fopen(file, "r");
+a = getc(fp);
+fclose(fp);
```
JimothyGreene, line 11: The reading of the strobelight status should be done on the Python side. This way you can test that both sides are working properly together.

BeVolt/Tests/Test_BSP_Strobelight.c
```diff
@@ -0,0 +1,22 @@
...
+printf("After Strobe_On, current status is %c", a);
+printf(" ... ");
+
+BSP_Strobe_Off();
+fp = fopen(file, "r");
+a = getc(fp);
+fclose(fp);
```
JimothyGreene, line 18: Same as above

BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
```diff
@@ -0,0 +1,19 @@
...
+    strobe = 0
+    os.makedirs(os.path.dirname(file), exist_ok=True)
+    with open(file, 'a') as csvfile:
+        pass   #creates file if it doesn't exist
+    with open(file 'r') as csvfile: #read file
+        csvreader = csv.reader(csvfile)
+        if len(strobe):
+            return int (strobe)
+        else:
+            return 0
```
JimothyGreene, line 19: You never actually update the value of strobe, so this won't tell us anything about the actual state of the strobelight.

AleNeff:
For some reason nothing would compile earlier when I had the simulate.py
file inside the BeVolt folder, I’ll take another look later / tomorrow and
try to figure out how to make that work so I can add testing for the .py

JimothyGreene:
Was it the compilation that was the issue? Or running the python script?

BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
AleNeff: I've looked at the other simulation branches, I have no idea what "adding a lock" is though..

BeVolt/Tests/Test_BSP_Strobelight.c
AleNeff: So what's the point of the C test file then?

AleNeff:
both compilation / running the python were issues. I think there's something wrong with my directory, it's a more far-reaching problem than just with this specific issue. When do you think you'd be available for a zoom call? That'd make it was easier to figure this stuff out because I'm still pretty new to this

BeVolt/Tests/Test_BSP_Strobelight.c
JimothyGreene: The C file should set the different states of the strobe light. Just as if the actual microcontroller code would do in certain situations. 

The python file would then act as the physical hardware, displaying an output for us to visually see.

BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
JimothyGreene: The point is to read the value that’s written in the CSV file. 

Right now, you set strobe to 0, do some stuff with a CSV file that ultimately doesn’t affect anything, then return int(strobe) which will always be 0 since you never actually read the value in csvreader no to the strobe variable.

BeVolt/BSP/Simulator/DataGeneration/Strobelight.py
```diff
@@ -0,0 +1,22 @@
...
+        pass   #creates file if it doesn't exist
+    with open(file, 'r') as csvfile: #read file
+        csvreader = csv.reader(csvfile)
+        for row in csvreader:
+            strobe.append(row)
+        if len(strobe):
+            return int(strobe[0][0])
+        else:
+            return 0
+print(read())
```
JimothyGreene, line 22: Can remove this now

BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
JimothyGreene: I would go ahead and change this to "BSP_Strobelight.h"

BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
JimothyGreene: Locking the file is just done like this:

After you open a file, put the following lines after
```
int fno = fileno(fp);
flock(fno, LOCK_EX);
```
And then before you close the file:
```
flock(fno, LOCK_UN);
```
And you'll have to include this header: ```<sys/file.h>```

AleNeff:
okay, I think I fixed those last few things. I'm about to leave my house, I'll try to tune in for the meeting later and I'll commit/push this stuff later tonight

BeVolt/BSP/Simulator/Src/BSP_Strobelight.c
```diff
@@ -0,0 +1,52 @@
...
+static const char* file = "BSP/Simulator/DataGeneration/Data/Strobe.csv";
+int StrobeStatus = 0;
+
+/**
+ * @brief   Create text file if none exists (Simulator)
+ * @param   None
+ * @return  None
+ */
+void BSP_Strobe_Init(void) {
+	FILE* fp = fopen(file, "w+");
```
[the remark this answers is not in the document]
AleNeff: No reason, I'll make it a "w"

BeVolt/Tests/Test_BSP_Strobelight.c
```diff
@@ -0,0 +1,22 @@
+#include "BSP_Strobelight.h"
+#include <stdio.h>
+
+
+int  main(){
+char strobeIn;
```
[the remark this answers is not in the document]
AleNeff: gotcha

AleNeff:
I changed those last two things, should this be ready to be merged w master?

PR #109, opened 2020-07-03
Simulator file locks

SijWoo:
Locks have been added to the BSP C src files and python scripts.

**NOTE:** BSP_Strobelight.c and Strobelight.py are not using the config path variables for the file location. Need to fix that in separate branch and pull request.

### DIFF
--- a/BSP/Simulator/DataGeneration/simulate.py
+++ b/BSP/Simulator/DataGeneration/simulate.py
@@ -123,7 +123,12 @@ def main():
     if input() == 'start':
         # Initial capacity is (2500*14)/(2950*14)=0.847 i.e. 84.7% charged
         init_capacity_mah = 2500 * config.num_batt_cells_parallel_per_module
-        BeVolt = battery.Battery(30, config.total_batt_pack_capacity_mah, init_capacity_mah)
+
+        # Amperes current draw of the electrical system
+        ampere_draw = 30
+
+        # Create state of the battery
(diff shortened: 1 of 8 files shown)


BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -120,7 +120,7 @@ def main():
     print("Type 'start' to start BeVolt. Otherwise, you can specify the types of data to simulate.")
     print(">>", end="")
     if input() == 'start':
-        BeVolt = battery.Battery(30, 2950*434, 2900*434)
+        BeVolt = battery.Battery(30, 2950*434, 2500*434)
```
chaseblock, line 123: It probably wouldn't be a bad idea to give these some variable names.

SijWoo: The first parameter is the amperes current that the electrical system will be drawing. The second parameter is supposed to be the total battery pack's capacity in mAh. The third parameter is the initial capacity of the battery pack you want to start out at in mAh. I lowered the initial capacity to 2500 because the voltage level at 2900 makes the BPS have an overvoltage fault. 

2950mAh is the capacity of each lithium ion cell. The 434 is the number of individual cells in the entire battery pack. I just realized the 434 is not correct because that assumes 434 cells are in parallel when it's supposed to be 14 cells in parallel (so 2950*14).

SijWoo: I added variables in config.py for those parameters. Let me know if there's anything else I can improve

PR #112, opened 2020-07-18
Heartbeat

AleNeff:
I believeeee this issue should be finished now, if there's formatting problems or something I can fix that

### DIFF
--- a/Apps/Src/main.c
+++ b/Apps/Src/main.c
@@ -73,6 +73,7 @@ int main(){
 		}
 
 		// TODO: Implement heartbeat for RUN light at a visible frequency
+		BSP_Light_Toggle(RUN);
 
 		// Update necessary
 		// CAN_SendMessageStatus()	// Most likely need to put this on a timer if sending too frequently
--- a/Apps/Src/main.c
+++ b/Apps/Src/main.c
@@ -145,7 +145,7 @@ void preliminaryCheck(void){
 void heartbeat(void){
 	// increment heartcount variable once per while(1) loop
 	heartcount = (heartcount + 1)%(heartbeatDelay);
-	if(heartcount = 0) {
+	if(heartcount == 0) {
 		BSP_Light_Toggle(RUN);
 	}
 }
--- a/Apps/Src/main.c
+++ b/Apps/Src/main.c
@@ -143,7 +143,7 @@ void preliminaryCheck(void){
 void heartbeat(void){
 	// increment heartcount variable once per while(1) loop
 	static int heartcount;
-	heartcount = (heartcount + 1)%(heartbeatDelay);
+	heartcount = (heartcount + 1)%(HEARTBEAT_DELAY);
 	if(heartcount == 0) {
 		BSP_Light_Toggle(RUN);
 	}
--- a/Apps/Src/main.c
+++ b/Apps/Src/main.c
@@ -16,11 +16,11 @@
 #include "BSP_Lights.h"
 #include "BSP_WDTimer.h"
 
-
 cell_asic Minions[NUM_MINIONS];
 bool override = false;		// This will be changed by user via CLI
 char command[COMMAND_SIZE];
 
+void heartbeat(void);
 void initialize(void);
 void preliminaryCheck(void);
 void faultCondition(void);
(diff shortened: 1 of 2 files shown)


chaseblock:
Should this be targeting the master or simulator branch?

AleNeff:
Ahh sorry I meant to target the simulator branch

chaseblock:
All good. Fixed.

Apps/Src/main.c
```diff
@@ -137,6 +139,17 @@ void preliminaryCheck(void){
 	}
 }
 
+/** heartbeat
+ * Toggle heartbeat at visible frequency (RUN light)
+ */
+void heartbeat(void){
+	// increment heartcount variable once per while(1) loop
+	heartcount = (heartcount + 1)%(heartbeatDelay);
+	if(heartcount = 0) {
```
chaseblock, line 148:
```suggestion
	if(heartcount == 0) {
```

Apps/Src/main.c
```diff
@@ -72,7 +74,7 @@ int main(){
 			break;
 		}
 
-		// TODO: Implement heartbeat for RUN light at a visible frequency
+		heartbeat();
```
ClarkPoon, line 77: When I compile, I get "warning: implicit declaration of function ‘heartbeat’ [-Wimplicit-function-declaration]". You need to add a function prototype for heartbeat() to the top of main.c, so the compiler knows that this function is defined.

Apps/Src/main.c
```diff
@@ -44,6 +44,8 @@ int main(){
 
 	BSP_WDTimer_Start();
 
+	uint32_t heartcount = 0;
```
ClarkPoon, line 47: When I compile I get " warning: unused variable ‘heartcount’ [-Wunused-variable]" because heartcount is declared as local to main().

Apps/Src/main.c
```diff
@@ -137,6 +139,17 @@ void preliminaryCheck(void){
 	}
 }
 
+/** heartbeat
+ * Toggle heartbeat at visible frequency (RUN light)
+ */
+void heartbeat(void){
+	// increment heartcount variable once per while(1) loop
+	heartcount = (heartcount + 1)%(heartbeatDelay);
```
ClarkPoon, line 147: When I compile this, I get "error: ‘heartcount’ undeclared (first use in this function);" This is because heartcount is local to main(). This can be fixed by declaring heartbeat as a static variable inside of heartbeat().

Config/Inc/config.h
```diff
@@ -54,4 +54,9 @@ typedef enum {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE = 3} SafetySta
 // Define how many temperature sensors are connected to each board
 #define MAX_TEMP_SENSORS_PER_MINION_BOARD	16
 
+//--------------------------------------------------------------------------------
+// HeartBeat Delay Ticks
+// Define heartbeatDelay as # of desired while(1) loops per toggle
+#define heartbeatDelay 0x6000
```
ClarkPoon, line 60: Can you make this all caps? Generally we make #defines all caps and variables not all caps, so it is easier to tell them apart at a glance.

ClarkPoon: Additionally, this value should be much smaller for the run light to toggle in a reasonable amount of time (I got this working with 0x60, but you can experiment to find a value that you think is best)

Apps/Src/main.c
```diff
@@ -16,11 +16,11 @@
...
 
-
 cell_asic Minions[NUM_MINIONS];
 bool override = false;		// This will be changed by user via CLI
 char command[COMMAND_SIZE];
 
+void heartbeat(void);
 void initialize(void);
 void preliminaryCheck(void);
 void faultCondition(void);
```
chaseblock, line 26: We haven't really been good at this, but I realized that all these functions could probably benefit from being marked `static`.

Config/Inc/config.h
chaseblock: Out of curiosity, does the simulator run at real-time, or is it faster than what we would expect on the board?

PR #115, opened 2020-07-22
Timer Simulation

sugamarora99:


BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -44,6 +48,8 @@ def display(battery=None):  #print watchdog countdown
     global frequency 
     frequency = PLL.Get_Frequency()
 
+    #Timer.Tick()
```
ClarkPoon, line 51: Can you remove this code if we are not using it?

PR #124, opened 2020-07-28
Add 'PLL' and 'CAN' to both simulator menus

ClarkPoon:
Made 'PLL' and 'CAN' valid menu options regardless of if the simulator is using a battery object or by specifying a configuration (charging vs discharging and low/normal/high values).

### DIFF
--- a/BSP/Simulator/DataGeneration/simulate.py
+++ b/BSP/Simulator/DataGeneration/simulate.py
@@ -200,58 +200,37 @@ def main():
         except KeyboardInterrupt:
             curses.endwin()
             if BeVolt is not None:
-                print("\n\rWould you like to change 'wires', 'quit', 'PLL', or send a CAN message ('CAN')?")
-                print(">>", end="")
-                choice = input()
-                if choice == 'wires':
-                    change_wires(BeVolt)
-                    stdscr = curses.initscr()
-                    curses.start_color()
-                elif choice == 'quit':
-                    break
-                elif choice == 'PLL':
-                    print("Enter the frequency you would like to change the clock to in Hz.")
-                    frequency = int(input())
-                    PLL.Change_Frequency(frequency)
-                elif choice == 'CAN':
-                    print("Enter the CAN ID for the system you wish to simulate. Leave out '0x'.")
-                    id = input()
-                    while(CAN.Invalid_CAN_ID(id) == True):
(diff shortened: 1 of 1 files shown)


BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -200,7 +200,7 @@ def main():
         except KeyboardInterrupt:
             curses.endwin()
             if BeVolt is not None:
-                print("\n\rWould you like to change 'wires', 'quit', or 'PLL'?")
+                print("\n\rWould you like to change 'wires', 'quit', 'PLL', or send a CAN message ('CAN')?")
```
chaseblock, line 203: This choice menu is getting a little long. Probably fine for now, but if much more is to be added, I suggest doing a menu sort of thing.
```
Select a #: 
1. Wires
2. PLL
3. ...
```

BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -213,18 +213,32 @@ def main():
...
+                    while(CAN.Invalid_CAN_ID(id) == True):
+                        print("Invalid CAN ID. Try again.")
+                        id = input()
+                    print("Enter up to 8 bytes of the CAN message that you would like to send, and separate each byte by a ','. Leave out '0x'.")
+                    message = input().split(',')
+                    CAN.Send_Message(id, message, len(message))
                 else:
                     print("That is not a valid option. Continuing simulation...")
                     stdscr = curses.initscr()
                     curses.start_color()
```
chaseblock, line 228: So at this point, everything from line 210 to 228 is exactly the same as what is repeated below it. I think it would probably be better to restructure this so that we don't have to repeat code as much. Maybe something like this.

```python
if BeVolt is not None:
    print('wires, PLL, CAN, quit')
else:
    print('config, PLL, CAN, quit')
print('>>')

if choice == 'wires' and BeVolt is not None:
    # Do stuff
elif choice == 'quit':
    # ...
elif choice == 'PLL':
    # ...
elif choice == 'CAN':
    # ...
elif choice == 'config' and BeVolt is None:
    # ...
else:
    # invalid
```

BSP/Simulator/DataGeneration/simulate.py
ClarkPoon: Good idea. I think I might as well implement this now.

PR #125, opened 2020-07-29
Add i2 c

AleNeff:
EEPROM can be read from ctrl-c menu in simulator - errors are return as 0x__ values rather than error names as of right now.

### DIFF
--- a/BSP/Simulator/DataGeneration/I2C.py
+++ b/BSP/Simulator/DataGeneration/I2C.py
@@ -4,10 +4,18 @@ import fcntl
 import config
 
 file = config.directory_path + config.files['I2C']
+# dump EEPROM all faults LIKE serialprintdata in eeprom file
+# read specific addresses out SIMILAR TO BSP_I2C_READ
 
 """
 @brief output the EEPROM error messages
     Function to be called by simulate.py
 """
 
-def 
+def EEPROM_Dump():
+    #Dump all faults from I2C.csv file
+    pass
+
+def I2C_Read(startAddress):
+    #Read faults from I2C.csv starting at startAdddress
+    pass
--- a/BSP/Simulator/DataGeneration/I2C.py
+++ b/BSP/Simulator/DataGeneration/I2C.py
@@ -8,14 +8,14 @@ file = config.directory_path + config.files["I2C"]
 # read specific addresses out SIMILAR TO BSP_I2C_READ
 
 faultDict = {
-    "0x01" : "FAULT_HIGH_TEMP",
-    "0x02" : "FAULT_HIGH_VOLT",
-    "0x04" : "FAULT_LOW_VOLT",
-    "0x08" : "FAULT_HIGH_CURRENT",
-    "0x10" : "FAULT_WATCHDOG",
-    "0x20" : "FAULT_CAN_BUS",
-    "0x40" : "FAULT_VOLT_MISC",
-    "0x80" : "DATA_SOC"
+    "0x01" : "High Temp",
+    "0x02" : "High Volt",
+    "0x04" : "Low Volt",
+    "0x08" : "High Current",
+    "0x10" : "Watchdog",
+    "0x20" : "Can Bus",
+    "0x40" : "Misc Volt",
+    "0x80" : "Charge"
 }
 
 errornames = bool
@@ -26,7 +26,6 @@ errornames = bool
 """
 
 def EEPROM_Dump(errornames):
-    Dump = []
     if os.stat(file).st_size != 0:
         os.makedirs(os.path.dirname(file), exist_ok=True)
         with open(file, 'r') as csvfile:
@@ -34,26 +33,27 @@ def EEPROM_Dump(errornames):
             csvreader = csv.reader(csvfile)
             DataList = list(csvreader)
             if errornames == 1:
+                Dump = ""
                 for fault in DataList:
                     if fault == TERMINATOR:
                         break
                     fault = fault[0]
                     fault = faultDict[fault]
-                    Dump.append(fault)
+                    Dump += fault
+                    Dump += ", "
                 fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
                 return Dump
             else:
+                Dump = []
                 for fault in DataList:
                     if fault == TERMINATOR:
                         break
                     Dump.append(fault)
                 fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
                 return str(Dump)
-            return "No EEPROM data"
     return "No EEPROM data"
 
 def I2C_Read(startAddress, errornames):
-    ReadFaults = []
     if os.stat(file).st_size != 0:
         os.makedirs(os.path.dirname(file), exist_ok=True)
         with open(file, 'r') as csvfile:
@@ -61,6 +61,7 @@ def I2C_Read(startAddress, errornames):
             csvreader = csv.reader(csvfile)
             Faults = list(csvreader)
             if errornames==1:
+                ReadFaults = ""
                 for i in range(startAddress, len(Faults)):
                     if Faults[i] == ["0x00"]:
                         return "This value is too high. Nothing in the EEPROM at this address yet."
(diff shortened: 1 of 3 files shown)


BSP/Simulator/DataGeneration/I2C.py
```diff
@@ -0,0 +1,62 @@
...
+
+def I2C_Read(startAddress):
+    ReadFaults = []
+    if os.stat(file).st_size != 0:
+        os.makedirs(os.path.dirname(file), exist_ok=True)
+        with open(file, 'r') as csvfile:
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)    
+            csvreader = csv.reader(csvfile)
+            Faults = list(csvreader)
+            for i in range(startAddress, len(Faults)):
```
ClarkPoon, line 52: While this is valid and how you would do it in C, I think something like
```
for fault in Faults:
    if fault == 0:
        . . .
```
is more readable.

AleNeff:
Thanks for all the feedback, I’ll make as many of these changes as I can
today and get it to a cleaner final state

BSP/Simulator/DataGeneration/I2C.py
AleNeff: So for some reason, each element in the Faults[] list is actually a "[0x__]" as opposed to just being an integer value

BSP/Simulator/DataGeneration/I2C.py
```diff
@@ -0,0 +1,62 @@
+import csv
+import os
+import fcntl
+import config
+
+TERMINATOR = ["0xff"]
```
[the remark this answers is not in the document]
AleNeff: ["0xff"] is what is actually found at each index of the Faults array -- it does seem very weird & took a while to realize, I'm still not sure why it's like that.

BSP/Simulator/DataGeneration/I2C.py
```diff
@@ -0,0 +1,62 @@
...
+        os.makedirs(os.path.dirname(file), exist_ok=True)
+        with open(file, 'r') as csvfile:
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)    
+            csvreader = csv.reader(csvfile)
+            DataList = list(csvreader)
+            for fault in DataList:
+                if fault == TERMINATOR:
+                    break
+                # print faultDict[str(fault).strip('[]')]
+                # I'd like to printout each error's name but cannot get this dictionary printing to work
```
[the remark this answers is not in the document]
AleNeff: So since each element in the array is actually a "["0x##"], as in each one is for some reason its OWN 1 element list, this wasn't working. I mostly resolved everything else but there's a lot of efficiency-based changes I could make if I could figure out why this is happening. I'd like to implement the dictionary-based printout before this is merged to the master but I think I'd need some help understanding why each element of the Fault[] list is itself another list.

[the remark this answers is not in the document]
ClarkPoon: I think this is occurring because I2C.csv isn't a real .csv file (I think it was originally called EEPROM.txt or something like that), since each value in the file is separated by a newline character instead of a comma. Try setting the delimiter character to newline (I think this is how you would do it):
```
csvreader = csv.reader(csvfile, delimeter='\n')
```
For more information on the csv module, see:
https://docs.python.org/3/library/csv.html

BSP/Simulator/DataGeneration/I2C.py
ClarkPoon: Right, I think that Faults is a list of lists because of the issue I have described above. Once that is fixed, each element in Faults will be a string, not an int. This can be fixed either by comparing an element in Faults to a string like this:
```
for fault in Faults:
    if fault == '0x00':
        ....
```
or by converting the strings (which have base 16 numbers) into integers like this:
```
for fault in Faults:
    if int(fault, 16) == 0:
        ....
```

BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -23,6 +24,7 @@
 CANbox = None   #box that displays the CAN messages
 lights_names = ['EXTRA', 'CAN', 'WDOG', 'UVOLT', 'OVOLT', 'OTEMP', 'OCURR', 'RUN', 'FAULT']
 frequency = None
+maxEEPROMAddress = 0x1FFFF
```
ClarkPoon, line 27: It has just occurred to me that I converted from decimal to hex incorrectly, and 16KB would give a maximum address of 0x3FFF

BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -213,10 +215,46 @@ def main():
...
+                    else:
+                        print("Invalid entry...", end="\n")
+                        print("Defaulted to raw data.")
+                    if choiceEEPROM1 == 'all':
+                        print(I2C.EEPROM_Dump(errormsg))
+                        print("Enter to continue simulator:")
+                        print(">>", end="")
+                        choice = input()
+                    elif choiceEEPROM1 == 'read':
+                        print("Enter address to read faults from (in hex format).")
```
ClarkPoon, line 239: The idea I had was for the user to be able to specify a range of EEPROM addresses that they want to read from (The addresses don't necessarily have to be faults). For example, the user might want to see the values of the pointers that are stored in the EEPROM that keep track of how large each array of faults in the EEPROM is.

BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -213,10 +215,46 @@ def main():
...
+                    print("Enter 'all' for all data or 'read' to enter specific address to read.")
+                    print(">>", end="")
+                    choiceEEPROM1 = input()
+                    print("Enter 'raw' to read the raw hex values or 'msg' for the translated error messages.", end="\n")
+                    print("If invalid response is given, default is raw data.")
+                    choiceEEPROM2 = input()
+                    if choiceEEPROM2 == 'raw':
+                        errormsg = 0
+                    elif choiceEEPROM2 == 'msg':
+                        errormsg = 1
```
ClarkPoon, line 229: What is `errormsg` being used for? Is this just to keep track of whether the user chooses 'raw' or 'msg'?

BSP/Simulator/DataGeneration/simulate.py
```diff
@@ -213,10 +215,46 @@ def main():
...
+                        EEPROMAddress = int(choiceEEPROM3, 16)
+                        if EEPROMAddress > 0 and EEPROMAddress < maxEEPROMAddress:
+                            print(I2C.I2C_Read(EEPROMAddress, errormsg))
+                            print("Enter to continue simulator:")
+                            choiceEEPROM = input()
+                        else:
+                            print("Invalid address...", end="\n")
+                            print("Enter to continue simulator:")
+                            choiceEEPROM = input()
+                    else:
```
ClarkPoon, line 250: The way this is set up leads to some counterintuitive behavior. For example, if the user gives an invalid input to the first question, they will only get an "invalid entry" error message after they answer the second question. This is confusing because it seems like the "invalid entry" error message is referring to the most recent question. I think it would be better if either
1. The error message was displayed immediately after the invalid entry, or
2. The error message was more specific, so it is obvious that it refers to the first question.

BSP/Simulator/DataGeneration/simulate.py
AleNeff: Yeah, that's all

BSP/Simulator/DataGeneration/simulate.py
AleNeff: Got it, went with #2

BSP/Simulator/DataGeneration/simulate.py
AleNeff: So do I need to implement an option to read from a given address to an end address & just spit out all the raw data between those two?

ClarkPoon: Yes, that is what I had in mind.

BSP/Simulator/DataGeneration/simulate.py
ClarkPoon: Can you change `errormsg` to something more descriptive? The name seems to imply that the variable holds an error message

AleNeff: made it returnErrorCodes

BSP/Simulator/DataGeneration/I2C.py
```diff
@@ -0,0 +1,92 @@
...
+        os.makedirs(os.path.dirname(file), exist_ok=True)
+        with open(file, 'r') as csvfile:
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)    
+            csvreader = csv.reader(csvfile)
+            Faults = list(csvreader)
+            if errornames==1:
+                ReadFaults = ""
+                for i in range(startAddress, endAddress):
+                    # if Faults[i] == ["0x00"]:
+                    #     return "This value is too high. Nothing in the EEPROM at this address yet."
```
ClarkPoon, line 67: If we don't need this commented out code, can you remove it?

BSP/Simulator/DataGeneration/I2C.py
```diff
@@ -0,0 +1,92 @@
...
+                            ReadFaults += fault
+                            ReadFaults += ", "
+                        else:
+                            ReadFaults += "Invalid, "
+                return (ReadFaults)
+            else:
+                ReadFaults = []
+                for i in range(startAddress, endAddress):
+                    # if Faults[i] == ["0x00"]:
+                    #     return "This value is too high. Nothing in the EEPROM at this address yet."
```
ClarkPoon, line 84: If we don't need this commented out code, can you remove it?

BSP/Simulator/DataGeneration/I2C.py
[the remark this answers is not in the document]
ClarkPoon: I looked it up and Python's csv.reader() method is hard-coded to use either '\r' or '\n' as the end of line character, which is what is causing each entry in the EEPROM to be read as a separate list. I wouldn't worry about changing this.

PR #126, opened 2020-07-30
Update master with simulator branch

SijWoo:
Warnings and certain errors have been fixed in the simulator branch that doesn't just pertain to the simulator (e.g. fixes with the Voltage library and the LTC6811 driver) and code for the stm32 will soon have to be updated for the new board. This merge doesn't mean the simulator is finished though, just need an update.


BSP/Template/Src/BSP_Contactor.c
```diff
@@ -1,6 +1,11 @@
 #include "BSP_Contactor.h"
```
ClarkPoon, line 1: Why did we edit the template file?

Drivers/Src/LTC681x.c
```diff
@@ -273,7 +272,7 @@ void LTC681x_adcvax(
   md_bits = (MD & 0x02) >> 1;
   cmd[0] = md_bits | 0x04;
   md_bits = (MD & 0x01) << 7;
-  cmd[1] =  md_bits | ((DCP&0x01)<<4) + 0x6F;
+  cmd[1] =  md_bits | ((DCP&0x01)<<4) | 0x6F;
```
ClarkPoon, line 275: Why was this changed?

Drivers/Src/LTC681x.c
```diff
@@ -956,7 +955,8 @@ int8_t LTC681x_rdstat(uint8_t reg, //Determines which Stat  register is read bac
       }
       else if (reg == 2)
       {
-        parsed_stat = data[data_counter++] + (data[data_counter++]<<8);              //Each gpio codes is received as two bytes and is combined to
```
ClarkPoon, line 959: Looks like there was some interesting code in here...

Drivers/Src/LTC681x.c
```diff
@@ -1370,6 +1368,11 @@ int16_t LTC681x_run_adc_redundancy_st(uint8_t adc_mode, uint8_t adc_reg, uint8_t
 
 	for (int cic=0; cic<total_ic; cic++)
 	{ 
+
```
ClarkPoon, line 1371: Why was this added?

Drivers/Src/LTC681x.c
```diff
@@ -1478,11 +1481,15 @@ int16_t LTC681x_run_adc_redundancy_st(uint8_t adc_mode, uint8_t adc_reg, uint8_t
 				}
 			}
 		}
+
+        // Reset system_open_wire
+        ic[cic].system_open_wire = 0;
```
ClarkPoon, line 1486: Why was this added?

BSP/Template/Src/BSP_Contactor.c
SijWoo: Hmmm, not sure why either..

Drivers/Src/LTC681x.c
SijWoo: This was giving a compiler warning, I tested it to make sure nothing changed

Drivers/Src/LTC681x.c
SijWoo: lol true

Drivers/Src/LTC681x.c
SijWoo: The ltc driver that we got from the LTC website did not implement the open wire checking function that well. The function would return a 64 bit where each bit represents an open wire on a certain pin but the cell_asic struct had a system_open_wire field per LTC6811 that was not being used even though it was defined in the struct. Originally it was confusing to just get a 64 bit number and not knowing which IC the open wire belonged to, especially since the LTC6811s aren't measuring 12 batteries. Having each IC have its own open_wire variable is easier to work with since we know which IC the open wire belonged to and it'll be scalable.

Drivers/Src/LTC681x.c
SijWoo: Same reasoning as above

PR #130, opened 2020-08-02
Display simulator errors.

ClarkPoon:
Made it so that the Python part of the simulator will freeze and show the error message instead of crashing if there is an error. This way, if someone adds an error to the code, they can see where the error is.


ClarkPoon:
After looking at this some more, I have realized that my current solution does not catch indentation errors.

PR #132, opened 2020-08-17
Finished fan drivers with 8 speed settings.

manthanand:


Apps/Src/Fans.c
```diff
@@ -0,0 +1,92 @@
...
+        TIM8->CCR3 += (accel * 500);
+        break;
+    case3:
+        if ((TIM8->CCR4 + (accel * 500)) > 0 || (TIM8->CCR4 + (accel * 500)) < 4000)
+        TIM8->CCR4 += (accel * 500);
+        break;
+    default:
+        break;
+    }
+}
```
ClarkPoon, line 92: This function looks like it would be easy to rewrite so that you can just set a speed from 0 to 4000. Maybe something like
```
void FanSpeedChange(uint8_t fan, uint16_t speed){
    switch (fan)
    {
    case1:
        if ((speed >= 0) && (speed <= 4000))
            TIM8->CCR1 = speed;
    . . .
```

Apps/Src/Fans.c
```diff
@@ -0,0 +1,92 @@
...
+        Increase or decrease speed (1 or -1)
+*/
+void FanSpeedChange(uint8_t fan, int8_t accel){
+    //Range of pulse is 0-4000, change in increments of 500
+    //First check to make sure that change is within range of values
+    //Load new value into Compare and Capture Register
+    switch (fan)
+    {
+    case 1:
+        if ((TIM8->CCR1 + (accel * 500)) > 0 || (TIM8->CCR1 + (accel * 500)) < 4000)
```
ClarkPoon, line 74: TIM8->CCR1 is a uint32_t, and accel is an int8_t. I don't know if this behavior is defined by the C standard or not (this may be undefined behavior, which is bad), but when I test this (add a uint32_t to the product of an int8_t and 500) on gcc on my machine, the result behaves as an unsigned integer. More specifically, 
```
(((uint32_t) 0) + ((int8_t) -1) * 500 > 0);
```
evaluates to true (presumably because of underflow).

Another thing to note, is that the `||` should be `&&`

manthanand:
I fixed all the bugs. Should I return a value if the function fails for some reason?

Apps/Src/Fans.c
manthanand: Changed variable to uint32_t

ClarkPoon (approved):

PR #134, opened 2020-08-22
Battery Balancing API

sugamarora99:


Drivers/Inc/BatteryBalancing.h
```diff
@@ -0,0 +1,61 @@
...
+ * @param   ic array of ICs in the system
+ * @return  None
+ */
+void ClearDischargeBit(int Cell, uint8_t total_ic, cell_asic ic[]);
+
+void testDischarge(cell_asic ic[]);
+
+void testStopDischarge(cell_asic ic[]);
+
+void testTheirDischarge(cell_asic ic[]);
```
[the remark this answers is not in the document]
chaseblock: Still want to know whether these "test" functions are used at runtime or are meant for their own files (for verification purposes). @ClarkPoon @sugamarora99

Drivers/Inc/BatteryBalancing.h
```diff
@@ -0,0 +1,60 @@
...
+ * @param   ic array of ICs in the system
+ * @return  None
+ */
+void Balancing_ClearDischargeBit(int Cell, uint8_t total_ic, cell_asic ic[]);
+
+void testDischarge(cell_asic ic[]);
+
+void testStopDischarge(cell_asic ic[]);
+
+void testTheirDischarge(cell_asic ic[]);
```
chaseblock, line 58: Who is "they" in `testTheirDischarge`?

chaseblock, line 58: Additionally, if these files are in fact called at runtime, rather than being for verification, rename to match the convention.

chaseblock (dismissed):

BSP/Simulator/DataGeneration/SPI.py
```diff
@@ -127,4 +127,4 @@ def read():
                 row[i] = int(row[i])
             values.append(row)
         fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)    # Unlock file
-    return values
+    return values
```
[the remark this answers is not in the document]
sugamarora99: This was unintentional

Drivers/Inc/BatteryBalancing.h
sugamarora99: These functions are not meant for runtime, they are supposed to be quick tests to make sure the hardware worked. After we verify that the hardware is all good they are no longer needed.

sugamarora99: "They" refers to the battery modules

Drivers/Inc/BatteryBalancing.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "LTC6811.h"
+#include "Voltage.h"
+
+/**
+ * @brief   Finds lowest voltage amongst the 31 modules. Meant to be called only in Balancing_Balance()
+ * @param   None
+ * @return  the  
+ */
+static uint16_t Balancing_Init(void);
```
ClarkPoon, line 16: This seems like a misleading name for this function. The name `Balancing_Init()` seems like it should initialize the battery balancing hardware. Also, why is this in the header file if it is a helper function that is only called in `Balancing_Balance()` ?

Drivers/Src/BatteryBalancing.c
```diff
@@ -0,0 +1,78 @@
...
+#include "Voltage.h"
+#include "BatteryBalancing.h"
+#include "config.h"
+#include "Current.h"
+
+#define CHARGING_TOLERANCE 0
+
+static uint16_t Balancing_Init(void) {
+	uint16_t minVoltage = 0;
+	if (Current_CheckStatus(0) == 1){
```
ClarkPoon, line 12: `Current_CheckStatus()` returns if the current is safe or not. Use `Current_IsCharging()` to check if it is charging.

Drivers/Src/BatteryBalancing.c
```diff
@@ -0,0 +1,78 @@
...
+		return;//If charging, return
+	}
+	Voltage_UpdateMeasurements();
```
ClarkPoon, line 15: Why are we doing this in here? We should be using the most recent voltage measurements (they are stored in a buffer by Voltage.c. I think it would make more sense to either:
1. Add a function to Voltage.c to find the minimum voltage of all of the modules
2. Add a function to Voltage.c to access the buffer of stored voltage values (`VoltageVal`)
3. Use  `Voltage_GetModuleMillivoltage()` without calling `Voltage_UpdateMeasurements()` first
I am also open to other ways to do this, if you have a better idea

Drivers/Src/BatteryBalancing.c
```diff
@@ -0,0 +1,78 @@
...
+		if(Voltage_GetModuleMillivoltage(i) < minVoltage) {
+			minVoltage = Voltage_GetModuleMillivoltage(i);//Find minimum voltage
+		}
+	}
+	return minVoltage;
+}
+	
+void Balancing_Balance(cell_asic Minions[]){ 
+	uint16_t lowest = Balancing_Init(); //get lowest voltage 
+	Voltage_UpdateMeasurements();
```
ClarkPoon, line 27: We should not be updating Voltage measurements in the battery balancing functions

Drivers/Inc/BatteryBalancing.h
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * @brief   Loops through all 31 modules, sets discharge bits for any module if its voltage is too high, and clears discharge bits for any modules with voltages that are too low
+ * @param   Minions array of the ICs that the modules are connected to
+ * @return  None
+ */
+void Balancing_Balance(cell_asic Minions[]);
```
ClarkPoon, line 23: Is this the only function that should be called by the battery balancing task, or by a main loop? If so, we should only expose this function.

Drivers/Src/BatteryBalancing.c
```diff
@@ -0,0 +1,78 @@
...
+		else {//Clear discharge bit of module if it reaches minimum
+			uint8_t ICIndex;
+			uint8_t MNumber;
+			Balancing_GetICNumber(k, &ICIndex, &MNumber);//Get module number and IC Board
+			Balancing_ClearDischargeBit(MNumber , 1, &Minions[ICIndex]);//Clear discharge bit
+		}
+	}
+}
+
+void Balancing_ClearDischargeBit(int Cell, uint8_t total_ic, cell_asic *ic){
```
ClarkPoon, line 42: Can you add the comments describing each function from the header file to this file as well? (Especially if we remove some of these functions from the header file)

ClarkPoon (changes requested):

ClarkPoon (dismissed):

ClarkPoon (approved):

PR #135, opened 2020-08-24
Created Software FIFO

manthanand:
Created software FIFO which is 32 columns by 8 rows for a max 8 bytes of data. I'm not sure what to do in the if statement but this pull request is only for the Software FIFO.

### DIFF
--- a/BSP/STM32F413/Src/BSP_CAN.c
+++ b/BSP/STM32F413/Src/BSP_CAN.c
@@ -145,9 +145,10 @@ void CAN1_RX0_IRQHandler(void)
         rxFifo[rxFifoHead][i] = RxMessage.Data[i]; //store data in FIFO
     }
     rxFifoHead = (rxFifoHead + 1)%CAN_FIFO_SIZE; //increment FIFO Head
-    if ((rxFifoHead + 1)%CAN_FIFO_SIZE == rxFifoTail){ //If the head is right below the tail
-        rxFifoTail++;
-    }
+    //If the head is right below the tail, increment tail to prevent overflow
+    if ((rxFifoHead + 1)%CAN_FIFO_SIZE == rxFifoTail){
+        rxFifoTail = (rxFifoTail + 1)%CAN_FIFO_SIZE;
+    } 
     //If high priority message, standard CAN message, with 1 byte is sent
     if ((RxMessage.StdId == 0x001)&&(RxMessage.IDE == CAN_ID_STD) && (RxMessage.DLC == 1)){
         // TODO: do stuff
--- a/BSP/STM32F413/Src/BSP_CAN.c
+++ b/BSP/STM32F413/Src/BSP_CAN.c
@@ -144,7 +144,8 @@ void CAN1_RX0_IRQHandler(void)
     for(int i = 0; i < 8; i++){
         CANFIFO[CANFIFOHEAD][i] = RxMessage.Data[i]; //store data in FIFO
         (++CANFIFOHEAD)%CANFIFOMAX; //increment FIFO Head
-        }
+    }
+    //If high priority message, standard CAN message, with 1 byte is sent
     if ((RxMessage.StdId == 0x001)&&(RxMessage.IDE == CAN_ID_STD) && (RxMessage.DLC == 1)){
         // TODO: do stuff
         RxFlag = true;
--- a/BSP/STM32F413/Src/BSP_CAN.c
+++ b/BSP/STM32F413/Src/BSP_CAN.c
@@ -143,11 +143,14 @@ void CAN1_RX0_IRQHandler(void)
     CAN_Receive(CAN1, CAN_FIFO0, &RxMessage);
     for(int i = 0; i < 8; i++){
         rxFifo[rxFifoHead][i] = RxMessage.Data[i]; //store data in FIFO
-        (++rxFifoHead)%CAN_FIFO_SIZE; //increment FIFO Head
+    }
+    (++rxFifoHead)%CAN_FIFO_SIZE; //increment FIFO Head
+    if (rxFifoHead == rxFifoTail){
+        //set error status
     }
     //If high priority message, standard CAN message, with 1 byte is sent
     if ((RxMessage.StdId == 0x001)&&(RxMessage.IDE == CAN_ID_STD) && (RxMessage.DLC == 1)){
         // TODO: do stuff
-        RxFlag = true;
+        RxFlag = true; //There is data in Fifo
     }
 }
--- a/BSP/STM32F413/Src/BSP_CAN.c
+++ b/BSP/STM32F413/Src/BSP_CAN.c
@@ -2,10 +2,13 @@
 #include "stm32f4xx.h"
 
 #define CAN_MODE        CAN_Mode_Normal
+#define CANFIFOMAX  32 //most messages it can take before wrapping around
 
 static CanTxMsg TxMessage;
 static CanRxMsg RxMessage;
-
+uint8_t CANFIFO[32][8] = {0}; //initialize 32 by 8 array containing FIFO data 
+CANFIFOHEAD = 0; //row of head
+CANFIFOTAIL = 0; //row of tail
 static bool RxFlag = false;
 
 /**
@@ -125,10 +128,11 @@ uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
 uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
     if(RxFlag){
 		for(int i = 0; i < 8; i++){
-			data[i] = RxMessage.Data[i];
+			data[i] = CANFIFO[CANFIFOTAIL][i];
+            (++CANFIFOTAIL)%CANFIFOMAX; //increment FIFO tail
 		}
         *id = RxMessage.StdId;
-		RxFlag = false;
+        if (CANFIFOHEAD == CANFIFOTAIL) RxFlag = false; //If all data is taken, remove flag
 		return 1;
 	}
     return 0;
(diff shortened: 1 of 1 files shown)


BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -2,10 +2,13 @@
 #include "stm32f4xx.h"
 
 #define CAN_MODE        CAN_Mode_Normal
+#define CANFIFOMAX  32 //most messages it can take before wrapping around
 
 static CanTxMsg TxMessage;
 static CanRxMsg RxMessage;
-
+uint8_t CANFIFO[32][8] = {0}; //initialize 32 by 8 array containing FIFO data
```
chaseblock, line 9: Is this `32` related to `CANFIFOMAX`? If so just use that instead.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -2,10 +2,13 @@
 #include "stm32f4xx.h"
 
 #define CAN_MODE        CAN_Mode_Normal
+#define CANFIFOMAX  32 //most messages it can take before wrapping around
```
chaseblock, line 5: `CAN_FIFO_SIZE` would read a bit better.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -2,10 +2,13 @@
...
 
 static CanTxMsg TxMessage;
 static CanRxMsg RxMessage;
-
+uint8_t CANFIFO[32][8] = {0}; //initialize 32 by 8 array containing FIFO data 
+CANFIFOHEAD = 0; //row of head
+CANFIFOTAIL = 0; //row of tail
```
chaseblock, line 11: When declaring a variable, make sure to specify the type. In this case, `static int`.

Also, since these are not `const`, they should be named something like `rxFifoHead` and `rxFifoTail`.

BSP/STM32F413/Src/BSP_CAN.c
chaseblock, line 9: Since this isn't `const`, name it something like `rxFifo` instead, and make it static.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -137,7 +141,11 @@ uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
 void CAN1_RX0_IRQHandler(void)
 {
     CAN_Receive(CAN1, CAN_FIFO0, &RxMessage);
-
+    for(int i = 0; i < 8; i++){
+        CANFIFO[CANFIFOHEAD][i] = RxMessage.Data[i]; //store data in FIFO
```
chaseblock, line 145: If you are incrementing the fifo head each time, this will not work as you expect.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -137,7 +141,11 @@ uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
...
+        (++CANFIFOHEAD)%CANFIFOMAX; //increment FIFO Head
```
chaseblock, line 146: This does not do what you think.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -137,7 +141,11 @@ uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
...
+    }
```
chaseblock, line 147: Make sure you check that the FIFO is not full, or at least clearly specify what will happen if it is. Right now this would drop the least recent messge in favor of an incoming message. Is this what we want? Maybe @ClarkPoon could weigh in on this.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -125,10 +128,11 @@ uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
 uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
     if(RxFlag){
```
chaseblock, line 129: Do we need to check for an RxFlag? Is there some other way of determining whether our fifo is full or not?

BSP/STM32F413/Src/BSP_CAN.c
manthanand: I think I fixed this

BSP/STM32F413/Src/BSP_CAN.c
manthanand: I think this increments the row of the matrix and mods it by the max so it wraps around when done. I moved the incrementing of the head outside the for loop. Would that work or is there something else I am missing.

BSP/STM32F413/Src/BSP_CAN.c
manthanand: I am setting the flag whenever there is data, not when it is full. The flag will only go away once all the data is gone. It could also go away if 32 pieces of data arrive between the IRQ and CAN_Read.

chaseblock: Sorry, I used the wrong word there. I mean that we should be able to determine whether the fifo is non-empty by comparing the head and tail.

Marking the FIFO as empty when it overflows is probably not the behaviour we would want.

BSP/STM32F413/Src/BSP_CAN.c
chaseblock: Specifically, the two lines shown below are not equivalent.
```c
(++CANFIFOHEAD)%CANFIFOMAX;

CANFIFOHEAD = (CANFIFOHEAD + 1) % CANFIFOMAX;
```

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -137,9 +141,16 @@ uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
 void CAN1_RX0_IRQHandler(void)
 {
     CAN_Receive(CAN1, CAN_FIFO0, &RxMessage);
-
+    for(int i = 0; i < 8; i++){
+        rxFifo[rxFifoHead][i] = RxMessage.Data[i]; //store data in FIFO
+    }
+    (++rxFifoHead)%CAN_FIFO_SIZE; //increment FIFO Head
```
chaseblock, line 147: Better to have this outside the loop, but it still doesn't do what you think it does.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -125,10 +128,11 @@ uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
 uint8_t BSP_CAN_Read(uint32_t *id, uint8_t *data) {
     if(RxFlag){
 		for(int i = 0; i < 8; i++){
-			data[i] = RxMessage.Data[i];
+			data[i] = rxFifo[rxFifoTail][i];
+            (++rxFifoTail)%CAN_FIFO_SIZE; //increment FIFO tail
 		}
```
chaseblock, line 133: Why increment the tail from inside the loop? If you have 8 messages in the buffer, you're going to read a little bit of all of them.

BSP/STM32F413/Src/BSP_CAN.c
manthanand: I think I did this

BSP/STM32F413/Src/BSP_CAN.c
manthanand: I think I fixed this

BSP/STM32F413/Src/BSP_CAN.c
manthanand: Moved outside loop

BSP/STM32F413/Src/BSP_CAN.c
manthanand: I did this using an error status variable and comparing the locations of the head and tail. I think it will work but there might be something wrong with the logic.

BSP/STM32F413/Src/BSP_CAN.c
manthanand: Changed this everywhere I did it.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -123,23 +127,32 @@ uint8_t BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
...
 		for(int i = 0; i < 8; i++){
-			data[i] = RxMessage.Data[i];
+			data[i] = rxFifo[rxFifoTail][i];
 		}
+        rxFifoTail = (rxFifoTail + 1)%CAN_FIFO_SIZE; //increment FIFO tail
         *id = RxMessage.StdId;
-		RxFlag = false;
 		return 1;
+        //if the head is not right below the tail, remove error status
+        if ((rxFifoHead + 1)%CAN_FIFO_SIZE != rxFifoTail) rxFifoOverFlow = 0;
```
chaseblock, line 139: If you overflow by more than one message, I'm pretty sure that you'll lose all of the previous messages this way. Consider the following example.
```
-----
| a | <-- TAIL
| b |
| c |
| d |
| e | <-- HEAD
-----          overflow = 0
```
Say two more messages come in. Then your buffer will look like this:
```
-----
| f | <-- TAIL
| g | <-- HEAD
| c |
| d |
| e |
-----          overflow = 1
```
Now say that you read one message. Then you get something like this:
```
-----
| f | <-- HEAD, TAIL
| g |
| c |
| d |
| e |
-----          overflow = 0
```
Notice how `c`, `d`, `e`, and `f` will never be read, despite still being valid.

The proper way to handle overflow would probably be not with a flag, but rather by erasing the oldest element, i.e. by increment the tail by one before adding the element if the buffer is full. This way, only the oldest messages are lost, rather than messages that we may still want to attend to.

manthanand: I'm kind of confused by the FIFO you drew. If the head is storing data and moving down, and the tail is storing data and moving down, wouldn't that mean that there isn't any data in g,c,d,e? Also, I set the error bit when the head is right before the tail, not when the tail is right before the head so the overflow bit wouldn't be set in the second graph. If I'm reading it wrong then let me know. Basically what I was trying to do was set apart two events. One when the head == tail and the FIFO is empty or when head == tail and the FIFO is full (that's what I used the error bit for). Also, I am storing/reading data, then moving so I'm not sure if there would be data in F.

manthanand: If you want, I can just erase the oldest element by incrementing TAIL when there is overflow. It's definitely easier than this.

ClarkPoon:
This was solved by #226

PR #169, opened 2020-08-29
Simulator fans

manthanand:


ClarkPoon:
Can you add some way to interact with the fans in the simulator? Also, I rebased from simulator, so you have the updated menu

.vscode/settings.json
```diff
@@ -14,7 +14,6 @@
         "math.h": "c",
         "bsp_adc.h": "c"
     },
-    "python.pythonPath": "/usr/bin/python3",
```
ClarkPoon, line 17: Was this on purpose?

manthanand:
Would I do that in Fans.py?

BSP/Simulator/DataGeneration/Fans.py
```diff
@@ -0,0 +1,27 @@
...
+import fcntl
+import config
+
+#This module will hold the statuses of the Fans on the BPS.
+
+#path name to file
+file = config.directory_path + config.files['Fans']
+
+#returns integer representing duty cycle
+def read():
```
ClarkPoon, line 12: Can you call this somewhere in simulate.py, so we can see the status of the fans when we are running the simulator?

BSP/Simulator/DataGeneration/Fans.py
```diff
@@ -0,0 +1,27 @@
...
+    with open(file, 'a') as csvfile:
+        pass    # creates file if not exists
+    with open(file, 'r') as csvfile: #read file
+        fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)    # Lock file
+        csvreader = csv.reader(csvfile)
+        for row in csvreader:
+            duty.append(row)
+        fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)    # Unlock file
+        if len(duty):
+            return int(duty[0][0])
```
ClarkPoon, line 24: Right now, this looks like it only gets the duty cycle of the first fan. We should have a way to read each fan's duty cycle in the simulator.

BSP/Simulator/Src/BSP_Fans.c
```diff
@@ -0,0 +1,62 @@
...
+ * @param   None
+ * @return  None
+ */
+void BSP_Fans_Init(void)
+{
+    // Creates file if none exists
+    FILE* fp = fopen(file, "w+");
+    int fno = fileno(fp);
+    flock(fno, LOCK_EX);
+    fprintf(fp, "0000");
```
ClarkPoon, line 19: Is each digit supposed to represent a fan? Since this is a csv file, I think it would make more sense if the duty cycles were separated by commas. This would also allow us to have more than ten duty cycle settings in the future.

BSP/Simulator/Src/BSP_Fans.c
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * @brief   Sets fan duty cycle
+ * @param   dutyCycle: int for duty cycle amount in range 0-8
+ *          fan: fan number whose speed should be changed (1-4)
+ * @return  None
+ */
+ErrorStatus BSP_Fans_Set(uint8_t fan, uint32_t dutyCycle){
+    //if input is outside of bounds
+    if (fan < 1 || fan > 4 || dutyCycle < 0 || dutyCycle > 8) return ERROR;
```
ClarkPoon, line 32: Can you split this into two lines, so it is easier to read?

ClarkPoon:
> Would I do that in Fans.py?

You would have to add it somewhere in simulate.py

.vscode/settings.json
manthanand: I only edited the c and h files so no.

BSP/Simulator/Src/BSP_Fans.c
```diff
@@ -0,0 +1,69 @@
...
+    if (dutyCycle < 0 || dutyCycle > MAX_SPEED) return ERROR;
+    FILE* fp = fopen(file, "r"); //Open file for reading
+    int fno = fileno(fp);  //Lock
+    flock(fno, LOCK_EX);    
+    fgets(Fans, ARRAY_SIZE, fp); //read from file
+    flock(fno, LOCK_UN); //Unlock
+    fclose(fp); //Close file
+    fp = fopen(file, "w"); //Open file for writing
+    //convert integer to string and store in Fans array
+    //NOTE: WHEN WE ADD DOUBLE DIGIT SPEED SETTINGS, REMOVE THE ZERO IN THE SPRINTF CALL!!!!!
```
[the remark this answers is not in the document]
manthanand: If I have to place commas in between the number, I think it has to be a character data type. So when I read the data into an array, it has to be %c. That's why I wrote 00,00... so it could store data with two digits.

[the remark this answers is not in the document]
manthanand: Never mind, it worked

manthanand:
I fixed it

ClarkPoon (approved):
Nice

PR #170, opened 2020-08-31
Set up RTD theme for Sphinx docs

shannonxtreme:
I set up the sphinx RTD theme. 
Installation instructions are available at: https://sphinx-rtd-theme.readthedocs.io/en/latest/installing.html

Basically, install sphinx, then: 
`pip install sphinx-rtd-theme`

Theme is added to source/conf.py with comments to show where and why. 

I am still looking into how to integrate RTD with the repo for continuous delivery of docs with new master merges.


shannonxtreme:
@ClarkPoon Getting Started has all the info you need to set up CI/CD for the docs: https://docs.readthedocs.io/en/stable/config-file/index.html

From the looks of it, you need to:
1. Add a config file and configure it
2. Set up a [Webhook](https://docs.readthedocs.io/en/stable/webhooks.html) 
3. Add [PDF as a downloadable format](https://docs.readthedocs.io/en/stable/downloadable-documentation.html) in the config file to automatically update the PDF version of your docs with each push

We can figure out styling etc. after this initial setup. I believe the final site is served at bps.readthedocs.io but I could be wrong. More information [here](https://docs.readthedocs.io/en/stable/custom_domains.html).

PR #200, opened 2020-09-12
changed the gpio pins for the leds

alekxoo:


ClarkPoon (changes requested):

ClarkPoon (commented):
Looks good to me. We will have to wait for hardware before we can test this though.

ClarkPoon (approved):

ClarkPoon (approved):

PR #201, opened 2020-09-19
Semaphores and mutexes

ClarkPoon:
added declarations for semaphores and mutexes listed in RTOS Shared Resources spreadsheet

### DIFF
--- a/.github/workflows/verify-build.yml
+++ b/.github/workflows/verify-build.yml
@@ -2,9 +2,7 @@ name: Verify Build
 
 on:
   push:
-    branches: [ master, simulator ]
   pull_request:
-    branches: [ master, simulator ]
 
 jobs:
   build:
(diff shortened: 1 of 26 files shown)


chaseblock:
As an aside: why are we keeping all the tasks in the same file? Would it make more sense to package, for example, our CLI task with the rest of the CLI?

I get that it makes it easier to give access to the semaphores to all of the tasks that may require them, but I think it would make more sense to just give visibility to the `VoltageBuffer_Mutex` to the tasks that `#include "Voltage.h"`.

Is the plan to modify the drivers to interface with the semaphores, or to just lock the threads until all the semaphores/mutexes are available for any drivers that they need to call?

SijWoo:
Usually I would have an app_<task name>.c file for each task but the main thing was the VoltTempMonitor task uses Voltage.h and Temperature.h

I think I misunderstood Clark's message earlier when I told him to put the semaphores inside Tasks.c. I meant the semaphores/mutex that are inter-task related, not the driver level semaphores/mutexes.

chaseblock (dismissed):

chaseblock (approved):

PR #202, opened 2020-09-20
Implemented MinionsASIC_Mutex

ClarkPoon:
I haven't been able to get the test to run in the simulator because it seg faults in `OSMutexPend()`


Drivers/Src/LTC6811.c
```diff
@@ -419,11 +499,27 @@ void LTC6811_set_discharge(int Cell, uint8_t total_ic, cell_asic ic[])
   {
     if (Cell<9)
     {
+      //take control of mutex
```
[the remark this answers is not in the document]
ClarkPoon: Ok. I will change that

SijWoo:
I just noticed the minions asic struct is read and modified in Temperature.c and Voltage.c, outside of the LTC6811 drivers. Should we place the mutexes in the Voltage and Temperature functions (one layer above the LTC6811 driver) instead of inside the LTC6811 driver?

Apps/Src/Temperature.c
```diff
@@ -108,10 +143,19 @@ ErrorStatus Temperature_ChannelConfig(uint8_t tempChannel) {
 		Minions[board].com.tx_data[4] = (AUX_I2C_NO_TRANSMIT << 4) + 0xF;
 		Minions[board].com.tx_data[5] = (0xF << 4) + AUX_I2C_NACK_STOP;
     }
+	//release mutex
+  	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
+  	assertOSError(err);
 
     // Send data
     wakeup_sleep(NUM_MINIONS);
```
chaseblock, line 151: Is the timing important on this command?

BSP/Simulator/Makefile
```diff
@@ -61,7 +61,8 @@ FLAGS = -Wall -g -std=c11 $(INC_DIR)
 ######################################
 # Defines
 ######################################
-DEF = -D_XOPEN_SOURCE=600 -DSIMULATION
+DEF = -D_XOPEN_SOURCE=700 -DSIMULATION
+DEF = -D_POSIX_C_SOURCE=200809L
```
chaseblock, line 65:
```suggestion
DEF += -D_POSIX_C_SOURCE=200809L
```

Drivers/Src/LTC6811.c
```diff
@@ -76,20 +77,20 @@ Copyright 2017 Linear Technology Corp. (LTC)
 /*********************************************************/
 /*** Code that was added by UTSVT. ***/
 /*********************************************************/
-static OS_MUTEX MinionsASIC_Mutex;
+OS_MUTEX MinionsASIC_Mutex;
 
 void LTC6811_Init(cell_asic *battMod){
+  //only create the mutex the first time this function is called (called by Voltage_Init() and Temperature_Init())
```
chaseblock, line 83: We only have one thread for the two right?

chaseblock (commented):
If it works it works.

That being said, is there any reason why we're using the mutexes in the Temperature/Voltage stuff, rather than the Driver level? I can think of the overhead of locking and unlocking as a disadvantage to changing to the driver layer, but it seems like it would be a bit easier to use.

ClarkPoon:
> If it works it works.
> 
> That being said, is there any reason why we're using the mutexes in the Temperature/Voltage stuff, rather than the Driver level? I can think of the overhead of locking and unlocking as a disadvantage to changing to the driver layer, but it seems like it would be a bit easier to use.

We moved the mutexes to the App level because Temperature.c and Voltage.c are directly modifying the minions array in some of their functions

Drivers/Src/LTC6811.c
ClarkPoon: Yes, that is correct.

chaseblock:
> We moved the mutexes to the App level because Temperature.c and Voltage.c are directly modifying the minions array in some of their functions

Oh I see. Well in that case it makes sense.

Apps/Src/Temperature.c
```diff
@@ -26,12 +30,25 @@ ErrorStatus Temperature_Init(cell_asic *boards){
 	wakeup_sleep(NUM_MINIONS);
 	LTC6811_Init(Minions);
 
+	//take control of mutex
+	OS_ERR err;
+  	OSMutexPend(&MinionsASIC_Mutex, 0, OS_OPT_PEND_BLOCKING, NULL, &err);
+  	assertOSError(err);
```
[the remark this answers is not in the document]
ClarkPoon: @SijWoo do you know if this is possible.

Also, what are the issues you are referring to @chaseblock ? The `assertOSError(err)` should put us into a fault state if an error is returned, which would mean that we don't read the minions array anymore.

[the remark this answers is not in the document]
chaseblock: Do we plan on reading sensor data in the fault condition (or at least trying)?

I suppose that since a fault should be the only thread running at that point, it's unreasonable to be checking mutexes. I was just talking about the possibility of entering some state where a mutex is held by some task that has been killed. Probably not actually a concern on second thought.

Apps/Src/Temperature.c
ClarkPoon: I am not sure. I can move the OSMutexPost() to before this line to be safe

chaseblock: Instead of doing that, might as well just not yield the mutex. This function call sits in between a Post() and Pend().

Apps/Src/Temperature.c
```diff
@@ -84,12 +106,25 @@ ErrorStatus Temperature_ChannelConfig(uint8_t tempChannel) {
...
     // Send data
     wakeup_sleep(NUM_MINIONS);
+	//take control of mutex
+  	OSMutexPend(&MinionsASIC_Mutex, 0, OS_OPT_PEND_BLOCKING, NULL, &err);
+  	assertOSError(err);
     LTC6811_wrcomm(NUM_MINIONS, Minions);
+	//release mutex
+  	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
+  	assertOSError(err);
     LTC6811_stcomm();
```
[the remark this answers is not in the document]
ClarkPoon: Ok, I will fix this

Apps/Src/Temperature.c
ClarkPoon: Ok

BSP/Simulator/Makefile
ClarkPoon: What does this do?

chaseblock: It adds...

Right now you are essentially doing (in pseudo-C):
```c
string DEF = "-D_XOPEN_SOURCE=700 -DSIMULATION"
string DEF = "-D_POSIX_C_SOURCE=200809L"
```
So at the end, DEF is just `"-D_POSIX_C_SOURCE=200809L"`.

If you append rather than reassign, it will end as `"-D_XOPEN_SOURCE=700 -DSIMULATION -D_POSIX_C_SOURCE=200809L"`, which is what I assume you meant to do.

Apps/Src/Temperature.c
```diff
@@ -116,10 +116,10 @@ ErrorStatus Temperature_ChannelConfig(uint8_t tempChannel) {
...
   	assertOSError(err);
     LTC6811_wrcomm(NUM_MINIONS, Minions);
+    LTC6811_stcomm();
 	//release mutex
   	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
   	assertOSError(err);
-    LTC6811_stcomm();
         
 	//take control of mutex
   	OSMutexPend(&MinionsASIC_Mutex, 0, OS_OPT_PEND_BLOCKING, NULL, &err);
```
chaseblock: Don't need to post and then immediately pend.

chaseblock (dismissed):

chaseblock (approved):

PR #203, opened 2020-09-20
Adding UART2 for CLI/BLE

manthanand:


manthanand:
The code will not compile since I changed the header file input parameters. I can change every call to my function to have the correct parameters since they will all be using USART 3. I am not sure if I should since the calls will be in other files.

SijWoo:
I was wondering, should we configure the BSP and drivers so we can use them in both RTOS and bare metal code? Like putting #ifdef statements over RTOS specific code. If we're planning on having the RTOS as our final version, then any place where we use the fifo like the one in UART can be using the Micrium fifo library @ClarkPoon @manthanand

manthanand:
When you say bare-metal code, does that mean just without an RTOS? I thought our only RTOS specific code was the actual engine running the RTOS (and other stuff like semaphores). Is the micrium fifo library being used elsewhere in our code (like is it something that we are already using and should start using everywhere or is it something that should be used in an RTOS for FIFO's)?

ClarkPoon:
> The code will not compile since I changed the header file input parameters. I can change every call to my function to have the correct parameters since they will all be using USART 3. I am not sure if I should since the calls will be in other files.

Can you change this so it builds?

ClarkPoon:
> When you say bare-metal code, does that mean just without an RTOS? I thought our only RTOS specific code was the actual engine running the RTOS (and other stuff like semaphores). Is the micrium fifo library being used elsewhere in our code (like is it something that we are already using and should start using everywhere or is it something that should be used in an RTOS for FIFO's)?

Yes, bare-metal code is when it is running without an RTOS. I don't think we are using the micrium fifo library anywhere right now, but I think the plan was for CAN to use it. @SijWoo correct me if I'm wrong on this.

SijWoo:
Since the purpose of the BSP is to make the application code hardware independent, would it make more sense to have an enum that lists out the different UART ports? Also we should do the same for BSP_SPI since the LTC and the new amperes board IC are on separate spi ports.
```
typedef enum {
    uart_usb,
    uart_ble
} uart_port_t;

typedef enum {
    spi_ltc6811,
    spi_as8510
} spi_port_t
```

SijWoo:
Yee that's correct, the CAN task is going to use the the micrium queue. To make the BSP_UART functions thread safe, we either need to make sure the fifo functions have mutexes or just use the micrium fifo. I think it would be easier to use the micrium fifo (queue) in the BSP_UART as well and not worry about testing the fifo functions for multi-thread safety.

SijWoo:
Also we need to be careful of reentrant functions with these. Using only local variables/input parameters or putting mutexes to global variables would prevent re-entrant issues.

manthanand:
I will try using the micrium fifo in the UART code. I will also try updating the simulator code to work with the the new header functions.

BSP/Simulator/Src/BSP_UART.c
```diff
@@ -50,104 +64,141 @@ void BSP_UART_Init(void) {
...
-
-        lineReceived = false;
-
-        pthread_mutex_unlock(&rx_mutex);
+        lineReceived3 = false;
+        pthread_mutex_unlock(&rx_mutex3);
+        return recvd;
+    }
+    if(lineReceived2 && (usart == UART_BLE)) {
+        pthread_mutex_lock(&rx_mutex2); //NEED TO CHANGE THIS **********
```
ClarkPoon, line 87: Do we still need to change this?

BSP/Simulator/Src/BSP_UART.c
```diff
@@ -50,104 +64,141 @@ void BSP_UART_Init(void) {
...
+        RxFifo_Peek(&data, UART_BLE);
+        while(!RxFifo_IsEmpty(UART_BLE) && data != '\r' && data != '\n') {
+            recvd += RxFifo_Get((uint8_t *)str++, UART_BLE);
+            RxFifo_Peek(&data, UART_BLE);
+        }
+        RxFifo_Get(&data, UART_BLE);
+        *str = 0;
+        lineReceived2 = false;
+        pthread_mutex_unlock(&rx_mutex2);//NEED TO CHANG THIS *********
```
ClarkPoon, line 96: Do we still need to change this?

BSP/Simulator/Src/BSP_UART.c
```diff
@@ -50,104 +64,141 @@ void BSP_UART_Init(void) {
...
     // Should not get here
     int retval = 1;
     pthread_exit(&retval);
 }
 
-static bool RxFifo_Get(uint8_t *data) {
-    if(!RxFifo_IsEmpty()) {
-        *data = rxBuffer[rxGet];
-        rxGet = (rxGet + 1) % RX_SIZE;
+static bool RxFifo_Get(uint8_t *data, UART_Port usart) {
```
ClarkPoon, line 156: Where are you using the `UART_Port usart` parameter? I think this function will behave incorrectly if you try to read from the UART_BLE, but the UART_USB fifo also has data in it.

BSP/Simulator/Src/BSP_UART.c
```diff
@@ -50,104 +64,141 @@ void BSP_UART_Init(void) {
...
         return true;
     }
     return false;
 }
 
-static bool RxFifo_Put(uint8_t data) {
-    if(!RxFifo_IsFull()) {
-        rxBuffer[rxPut] = data;
-        rxPut = (rxPut + 1) % RX_SIZE;
+static bool RxFifo_Put(uint8_t data, UART_Port usart) {
```
ClarkPoon, line 170: See comment for RxFifo_Get

BSP/Simulator/Src/BSP_UART.c
```diff
@@ -50,104 +64,141 @@ void BSP_UART_Init(void) {
...
+        rxPut2 = (rxPut2 + 1) % RX_SIZE;
         return true;
     }
     return false;
 }
 
-static bool RxFifo_Peek(uint8_t *data) {
-    if(!RxFifo_IsEmpty()) {
-        *data = rxBuffer[rxGet];
+static bool RxFifo_Peek(uint8_t *data, UART_Port usart) {
```
ClarkPoon, line 184: See comment for RxFifo_Get

BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -44,208 +67,598 @@ void BSP_UART_Init(void) {
...
     setvbuf(stdout, NULL, _IONBF, 0);
+    #ifdef RTOS
+    OSMutexCreate(&rx2Mutex, "RX 2 Mutex", &err);
+    OSMutexCreate(&tx2Mutex, "TX 2 Mutex", &err);
+    OSMutexCreate(&rx3Mutex, "RX 3 Mutex", &err);
+    OSMutexCreate(&tx3Mutex, "TX 3 Mutex", &err);
+    #endif
 }
-
+//***********THE FOLLOWING CODE IS FOR THE BM VERSION OF THE BPS***********
```
ClarkPoon, line 115: Can you spell it out as "bare metal" in the comments as well?

BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -44,208 +67,598 @@ void BSP_UART_Init(void) {
...
-
+        if(!TxFifo_Get((uint8_t *)&(USART3->DR),UART_USB)) USART_ITConfig(USART3, USART_IT_TC, RESET);
     }
+    if(USART_GetITStatus(USART3, USART_IT_ORE) != RESET);
+}
+
 
-    if(USART_GetITStatus(USART3, USART_IT_ORE) != RESET) {
+static bool TxFifo_Get(uint8_t *data, UART_Port usart) {
+    if(!TxFifo_IsEmpty(UART_USB) && (usart == UART_USB)) {
```
ClarkPoon, line 243: You should do something like this in the simulator version of these functions

BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -44,208 +67,598 @@ void BSP_UART_Init(void) {
...
+        RxFifo_Get(&data, UART_USB);
+        *str = 0;
+        OSMutexPend(&rx3Mutex, 0, OS_OPT_PEND_BLOCKING, &time, &err);
+        lineReceived3 = false;
+        OSMutexPost(&rx3Mutex, OS_OPT_POST_NONE, &err);
+        USART_ITConfig(USART3, USART_IT_RXNE, SET);
+        return recvd;
+    }
+    //lineRecieved2 is used by rx and tx but only rx mutex will be used for access
+    OSMutexPend(&rx2Mutex, 0, OS_OPT_PEND_BLOCKING, &time, &err);
```
ClarkPoon, line 382: What happens if the statement `if(lineReceived2 && (usart == UART_BLE))` is false and we return without posting the mutex? I think this could cause UART to freeze.

ClarkPoon (changes requested):
I didn't go super in-depth into BSP_UART.c for the real board because we will probably want to test that on the hardware

BSP/Simulator/Src/BSP_UART.c
manthanand: No, I left that there in an earlier version of my code.

manthanand:
Since I removed the #define at the start of the file, all of the code is commented out and that is causing the errors. If it is compiled with RTOS or BAREMETAL, it will work.

ClarkPoon (approved):

PR #204, opened 2020-09-25
Digitized amps

chaseblock:
Not yet been tested. Let me know what you think of the code.

It includes some modifications to SPI BSPs and adds an AS8510 driver for the new current board.

Requires the latest (not yet assembled) version of the leader board and the amperes board.


chaseblock:
The test is failing because I did not create a SPI3 BSP for the simulator, as I was informed that the SPI simulator was being redone and did not want to duplicate efforts.

SijWoo:
Since the BSP functions are meant to make the drivers hardware independent, having SPI3 and SPI1 in the BSP function names would make them sort of hardware dependent since those SPI ports are mapped for the STM32F413. I commented on this on the BSP_UART PR because the BLE module and USB use different UART ports. I suggested adding an input parameter to each BSP_UART function and having an enum like these to handle the different ports:
```
typedef enum {
    uart_usb,
    uart_ble
} uart_port_t;
```
so for the BSP SPI functions, we can have add the following enum as an input parameter for each BSP_SPI function:
```
typedef enum {
    spi_ltc6811,
    spi_as8510
} spi_port_t;
```

Though now I realize this can make the function definitions long. We should designate peripherals that use multiple ports with names related to drivers rather than port numbers. BSP_SPI_LTC6811 and BSP_SPI_AS8510 function names are good as well tbh

chaseblock:
> Since the BSP functions are meant to make the drivers hardware independent, having SPI3 and SPI1 in the BSP function names would make them sort of hardware dependent since those SPI ports are mapped for the STM32F413. I commented on this on the BSP_UART PR because the BLE module and USB use different UART ports. I suggested adding an input parameter to each BSP_UART function and having an enum like these to handle the different ports:

I agree that this would be good. I've gone ahead and added this to the SPI BSP here, at least on the microcontroller side. I am less familiar with how the simulator is laid out, so that remains unsupported right now for SPI3. The prototypes are there though.

Drivers/Inc/AS8510.h
```diff
@@ -0,0 +1,38 @@
...
+ * @param addr the address of the register
+ * @return the byte that was read
+ */
+uint8_t AS8510_ReadRegister(uint8_t addr);
+
+/* Gets the current from the AS8510
+ */
+int16_t AS8510_GetCurrent();
+
+#endif
```
ClarkPoon, line 38: you are missing a newline character

Drivers/Src/AS8510.c
```diff
@@ -0,0 +1,206 @@
...
+ */
+int16_t AS8510_GetCurrent() {
+    uint8_t dreg[2];
+
+    AS8510_ReadFromAddr(0, dreg, 2);
+
+    uint16_t reading = ((uint16_t)(dreg[0]) << 8) | (uint16_t)(dreg[1]);
+
+    return *((int16_t*)&reading); // TODO: how does this need to be modified to be useable?
+}
```
ClarkPoon, line 206: No newline

ClarkPoon (commented):
Some of your files are missing newlines at the end, and this still needs to be tested on hardware, but other than that, it looks good to me

manthanand (dismissed):

ClarkPoon (dismissed):

ClarkPoon:
@manthanand you Chase created different APIs for using SPI1 and SPI3 (see differences between his BSP_SPI.h and your BSP_SPI.h on the rtos branch). We need to make a decision on which one we like more before I can merge this.

manthanand:
I think it was sugam that worked on the SPI branch. I think we will have to use Chase's version since it accounts for multiple SPI's.

ClarkPoon:
> I think it was sugam that worked on the SPI branch. I think we will have to use Chase's version since it accounts for multiple SPI's.

Yes, you are correct that @sugamarora99 worked on that. I also prefer Chase's setup with an `enum` for the different spi ports better than just using functions called SPI or SPI3

ClarkPoon (approved):

Apps/Src/Amps.c
```diff
@@ -1,27 +1,43 @@
...
 #include "AS8510.h"
+#include "os.h"
+#include "Tasks.h"
+#include "BSP_SPI.h"
+
+static OS_ERR err;
+static CPU_TS ticks;
+static OS_MUTEX AmperesData_Mutex;
+
+static bsp_os_t spi3;
```
chaseblock, line 17: Is this used anywhere?

Apps/Src/Amps.c
```diff
@@ -1,27 +1,43 @@
...
  * battery pack.
  */
 
 #include "Amps.h"
 #include "AS8510.h"
+#include "os.h"
+#include "Tasks.h"
+#include "BSP_SPI.h"
+
+static OS_ERR err;
```
chaseblock, line 13: Is there an advantage to keeping this global rather than local? Seems like just another thing that we have to use locks to access...

Apps/Src/Amps.c
```diff
@@ -1,27 +1,43 @@
...
+static CPU_TS ticks;
```
chaseblock, line 14: Same comment as above.

Apps/Src/Amps.c
```diff
@@ -54,13 +74,65 @@ SafetyStatus Amps_CheckStatus(bool chargingOnly) {
  * @return 1 if charge, 0 if discharge
  */
 int8_t Amps_IsCharging(void) {
-	return latestMeasureMilliAmps < 0;
+	// TODO: Make sure that the amperes board is installed in such a way that negative => charging
+	OSMutexPend(&AmperesData_Mutex, 0, OS_OPT_PEND_BLOCKING, &ticks, &err);
```
chaseblock, line 78: There is no race condition here. No need for a lock.

Apps/Src/Amps.c
```diff
@@ -54,13 +74,65 @@ SafetyStatus Amps_CheckStatus(bool chargingOnly) {
...
+	return val;
 }
 
 /** Amps_GetReading
  * Measures the electrical current from the Ampere minion board
  * @return milliamperes value
  */
 int32_t Amps_GetReading(void) {
-	return latestMeasureMilliAmps;
+	OSMutexPend(&AmperesData_Mutex, 0, OS_OPT_PEND_BLOCKING, &ticks, &err);
```
chaseblock, line 91: No race condition, no need for a lock

Apps/Src/Amps.c
```diff
@@ -31,18 +47,22 @@ ErrorStatus Amps_UpdateMeasurements(void) {
...
+		assertOSError(err);
+		return SAFE;
+	}
+	else{
+		OSMutexPost(&AmperesData_Mutex, OS_OPT_POST_NONE, &err);
+		assertOSError(err);
+		return DANGER;
 	}
 
 	return DANGER;
```
chaseblock: Could simplify this all quite a bit by doing something like
```suggestion
	SafetyStatus stat = DANGER;
	OSMutexPend(&AmperesData_Mutex, 0, OS_OPT_PEND_BLOCKING, &ticks, &err);
	assertOSError(err);
	if((latestMeasureMilliAmps > MAX_CHARGING_CURRENT)&&(latestMeasureMilliAmps < MAX_CURRENT_LIMIT)&&(!chargingOnly)){
		stat = SAFE;
	}
	else if((latestMeasureMilliAmps <= 0)&&(latestMeasureMilliAmps > MAX_CHARGING_CURRENT)&&chargingOnly){
		stat = SAFE;
	}
	else{
		stat = DANGER;
	}

	OsMutexPost(&AmperesData_Mutex, OS_OPT_POST_NONE, &err);
	assertOSError(err);
	return stat;
```

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -1,8 +1,28 @@
...
+#ifdef RTOS
+#define SPI_Wait(SPIx)		if(((SPIx)->SR & (SPI_SR_TXE | SPI_SR_RXNE)) == 0 || ((SPIx)->SR & SPI_SR_BSY)){ \
+								if(SPIx == SPI1){	\
+									os->pend();	\
+								}	\	
+								else if(SPIx == SPI3){	\
+									os3->pend();	\
+								}	\
+							} 
+#endif
```
chaseblock: Maybe instead of `#define`s, we can use these things called `functions` where we don't have to break lines with `\`. If you really want a minor speed optimization, just declare them `inline`.

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -1,8 +1,28 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
 #include "BSP_SPI.h"
 #include "stm32f4xx.h"
+#include "os.h"
+#include "BSP_OS.h"
+
+static bsp_os_t *os;
+static bsp_os_t *os3;
```
chaseblock: I was trying to avoid things like this when I added the LUTs below. Let's avoid putting a whole bunch of `if`s everywhere if possible.

Apps/Src/Amps.c
ClarkPoon: I don't think so

Drivers/Src/AS8510.c
```diff
@@ -33,12 +33,48 @@
 #define AS8510_DEC_R2_V      0x46
 #define AS8510_FIR_CTL_REG_V 0x47
 
+static OS_SEM AmperesIO_Sem;
+static bsp_os_t spi_os;
+
+#ifdef RTOS
+void Amperes_Pend(){
```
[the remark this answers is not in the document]
ClarkPoon: This semaphore is used by the SPI3 ISR to signal when there is an interrupt (SPI message received from AS8510). I put it here because `BPS_SPI_Init()` uses this as a parameter. I suppose it might make more sense to put this in Amps.c and pass this as a parameter `AS8510_Init()`

Apps/Src/Amps.c
ClarkPoon: I will change this

Apps/Src/Amps.c
ClarkPoon: I will fix

Apps/Src/Amps.c
ClarkPoon: I will fix

Apps/Src/Amps.c
ClarkPoon: I will fix

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -52,6 +34,26 @@ static const uint16_t SPI_SELECT_PINS[NUM_SPI_BUSSES] = {
...
+#ifdef RTOS
+	if(((SPIx)->SR & (SPI_SR_TXE | SPI_SR_RXNE)) == 0 || ((SPIx)->SR & SPI_SR_BSY)){
+		if(SPIx == SPI1){
+			SPI_os[spi_ltc6811]->pend();
+		}
+		else if(SPIx == SPI3){
+			SPI_os[spi_as8510]->pend();
+		}
+	} 
+#endif
```
chaseblock:
```suggestion
#ifdef BAREMETAL
	while(((SPIx)->SR & (SPI_SR_TXE | SPI_SR_RXNE)) == 0 || ((SPIx)->SR & SPI_SR_BSY));
#endif

#ifdef RTOS
	if(SPIx == SPI1){
		SPI_os[spi_ltc6811]->pend();
	}
	else if(SPIx == SPI3){
		SPI_os[spi_as8510]->pend();
	} 
#endif
```

Apps/Src/Amps.c
```diff
@@ -0,0 +1,161 @@
...
+
+#include "Amps.h"
+#include "AS8510.h"
+#include "os.h"
+#include "Tasks.h"
+#include "BSP_SPI.h"
+
+static OS_MUTEX AmperesData_Mutex;
+
+static OS_SEM AmperesIO_Sem;
```
chaseblock, line 15: Needs init

chaseblock (commented):
Looks good to me.

ClarkPoon (dismissed):

ClarkPoon (approved):

PR #208, opened 2020-09-26
Faulttask

manthanand:


ClarkPoon:
@sugamarora99 can you review this?

Apps/Src/Task_FaultState.c
```diff
@@ -1,34 +1,65 @@
...
-
+    BSP_Contactor_Off();
     // Turn Strobe Light On
-
-    // Turn LEDs On
-
-    // Record State of BPS into EEPROM
-
+    // Turn LEDs On and logs Error into EEPROM
+    BSP_Light_Off(RUN); //turn of run light
```
sugamarora99, line 29: Typo

Apps/Src/Task_FaultState.c
```diff
@@ -1,34 +1,65 @@
...
+    CANbus_Send(TRIP, Message);
     // Push Contactor State message to CAN Q
-
+    data.b = 0;
+    CANbus_Send(CONTACTOR_STATE, Message);
+    char command[COMMAND_SIZE];
     while(1) {
-        // CLI
-        // WDOG Reset
+        if (BSP_UART_ReadLine(command)) CLI_Handler(command); // CLI
```
sugamarora99, line 62: Can you explain why we need CLI in two tasks? I'm not saying this code is incorrect, this was just a question I had.

manthanand: This is the only way I could think of running the CLI task and reset the watchdog at the same time since the CLI task doesn't do that automatically.

sugamarora99: Why do we have to run the CLI in this task?

manthanand: I'd assume it's so we can read the data that led to our fault state. There might be other reasons too, but that's probably the main one.

sugamarora99 (approved):

Apps/Src/Tasks.c
```diff
@@ -6,6 +6,10 @@
 #include "Voltage.h"
 #include "Temperature.h"
 
+#include "../../BSP/Inc/BSP_Lights.h"
```
ClarkPoon, line 9: Why are you including these here?

Apps/Src/Task_FaultState.c
ClarkPoon: We want to run the CLI task here so we can figure out what caused the fault condition. I think this is fine here because if we get stuck in the CLI because of mutexes, the watchdog timer will trip and clear all of the mutexes, and we will return to a fault condition, so the contactor will stay open. What do you think @chaseblock @SijWoo @sugamarora99 @manthanand ?

Apps/Src/Tasks.c
manthanand: So I can change the LED's that are on. It also lets me turn on the strobe light.

ClarkPoon: Do you use these anywhere in this file though? It looks like your only changes here are to include more files and to add commented out code

Apps/Src/Task_FaultState.c
ClarkPoon: I discussed with @chaseblock over lunch, and I think it would be a good idea to add an `#ifdef` so that we only enable the CLI in the fault condition if we are debugging (I don't think this should be included when the car is racing because it can be very intrusive if it causes the watchdog timer to trip).

Apps/Src/Tasks.c
manthanand: I thought this was the faultState file, not tasks file. I'll remove it.

manthanand: I just checked the commit history and Ethan made all those changes to tasks.c so I'm just going to undo all of them.

ClarkPoon: Ok, sounds good

ClarkPoon (approved):

PR #209, opened 2020-09-26
CLI Task for RTOS

sugamarora99:
CLI will pend for the resources needed for the specific command entered by the user. Also added "openwire" command to CLI that displays open wire data for the boards.


Apps/Src/Tasks.c
```diff
@@ -60,6 +61,65 @@ OS_SEM Fault_Sem4;
...
 
+
+/** Pend_ResourceForCLI
+ * Pends for desired resource that CLI needs
+ * @param resourceNum the resource you want to wait for, look at CLI_Resource enum in CLI.h
+ */
+void Pend_ResourceForCLI(int resourceNum){
+    switch(resourceNum){
+        case 0:
+            OS_Sem4Pend(&SafetyCheck_Sem4);
```
manthanand, line 72: Where is this function defined? I thought the function used for pending and posting was OS_SemPend and OS_SemPost (in os.h).

Apps/Src/Tasks.c
```diff
@@ -60,6 +61,65 @@ OS_SEM Fault_Sem4;
...
+            break;
+        case 1:
+            OS_MutexLock(&MinionsASIC_Mutex);
```
manthanand, line 75: Same for this

Apps/Src/Tasks.c
```diff
@@ -60,6 +61,65 @@ OS_SEM Fault_Sem4;
 cell_asic Minions[NUM_MINIONS];
 uint32_t WDog_BitMap = 0;
 
+
+/** Pend_ResourceForCLI
+ * Pends for desired resource that CLI needs
+ * @param resourceNum the resource you want to wait for, look at CLI_Resource enum in CLI.h
+ */
+void Pend_ResourceForCLI(int resourceNum){
```
manthanand, line 69: If only the CLI will be calling these functions, should they be in CLI.c instead of Tasks.c?

Apps/Src/Tasks.c
```diff
@@ -60,6 +61,65 @@ OS_SEM Fault_Sem4;
...
+        default:
+            break;
+    }
+}
+
+/** Post_ResourceForCLI
+ * Posts desired resource that CLI needs
+ * @param resourceNum the resource you want to post, look at CLI_Resource enum in CLI.h
+ */
+void Post_ResourceForCLI(int resourceNum){
```
manthanand, line 98: Should this also be in CLI.c?

Apps/Src/Tasks.c
```diff
@@ -227,23 +287,97 @@ void Task_BatteryBalance(void *p_arg) {
     }
 }
 
+void Check_ResourceForCLI(int resourceNum){
```
manthanand, line 290: What is the difference between this and Pend_ResourceForCLI?

Apps/Src/Tasks.c
sugamarora99: I made the Pend_ResourceForCLI() and Post_ResourceForCLI() functions because I thought all the semaphores/mutexes were going to be global variables declared in Tasks.c. I put these functions in Tasks.h so that way I could access the Task module's globals in the CLI module.

Apps/Src/Tasks.c
sugamarora99: It's the same function, I just renamed it from "Check_" to "Pend_". I will remove this.

Apps/Src/Tasks.c
manthanand: I think we could do this 2 ways. We could define the semaphores and mutexes in tasks.h and keep our functions as static in our specific c files (becuase if every thread had their own pend and post function in tasks.c it might get cluttered). Another idea could be to have these functions as just Post_Resource and Pend_Resource and any thread could call them to check these semaphores. What do you and @ClarkPoon think?

sugamarora99: I think we should do the latter.

manthanand: Regardless of if we do the second or the first one, I would recommend that the variable passed into these functions be the type of the enum that you are passing into the functions (rather than an int). I would also make each case in the switch case statement one of the enum types rather than an integer, just so it is easier to read.

Apps/Inc/CLI.h
```diff
@@ -69,6 +71,9 @@
 // Useful values
 #define NUM_TEMP_SENSORS_LAST_MINION    14
 
+//For pending/posting CLI resource semaphores/mutexes
+enum CLI_Resource{SafetyCheck, MinionsASIC, VoltageBuffer, TemperatureBuffer, OpenWireBuffer, AmperesData};
```
ClarkPoon, line 75: Can you make these all caps, so it is easy to see that these are constants when we are reading the code? Also, can you `typedef` this to something descriptive?

Apps/Inc/Tasks.h
```diff
@@ -51,6 +51,10 @@ void Task_CANBusConsumer(void *p_arg);
 
 void Task_BatteryBalance(void *p_arg);
 
+void Pend_ResourceForCLI(int resourceNum);
```
ClarkPoon, line 54: Can you add comments on how to use these functions?

Apps/Src/CLI.c
```diff
@@ -652,28 +654,45 @@ void CLI_Critical(void) {
...
+void CLI_OpenWire(void){
+	Voltage_OpenWireSummary();
+}
+
 /** CLI_All
  * Displays all information about BPS modules
- * (voltage, current, temperature, charge, contactor)
+ * (voltage, current, temperature, charge, contactor, open wires)
  */
 void CLI_All(void) {
```
ClarkPoon, line 668: The lower-level functions (Voltage, Temperature, etc.) should be pending and posting these mutexes. I know we already discussed at the meeting, so I am just posting this as a reminder.

Apps/Src/Tasks.c
ClarkPoon: We should be only pending and posting the mutexes in the functions that directly access the shared resource. In this case, you are calling a different function that does the actual access, so you should not pend or post the mutexes here (as we discussed at the meeting yesterday).

ClarkPoon (changes requested):

Issue #213, opened 2020-09-26
RTOS stack above BSP


SijWoo:
To make the Board Support Package support both bare metal and a RTOS, the team decided to make the BSP code RTOS independent. Each BSP src file that requires an OS shall take a `bsp_os_t` type input parameter in the Init function. This struct contains function pointers that are defined at an upper layer. If there is no OS, then the struct should be filled with dummy empty functions that does nothing.

Here is an example of the SPI peripheral using OS functions:
```
// BSP_OS.h
typedef struct {
    void (*pend)(void);
    void (*post)(void);
} bsp_os_t;
```

```
// LTC6811.c
bsp_os_t spi_os;
OS_SEM MinionsIO_Sem4;
void LTC6811_Pend(void) {
    CPU_TS ts;
    OS_ERR err;
    OSSemPend(&MinionsIO_Sem4,
                        0,
                        OS_OPT_PEND_BLOCKING,
                        &ts,
                        &err);
}

void LTC6811_Post(void) {
    OS_ERR err;
    OSSemPost(&MinionsIO_Sem4,
                        OS_OPT_POST_1,
                        &err);
}

void LTC6811_Init(void) {
    spi_os.pend = LTC6811_Pend;
    spi_os.post = LTC6811_Post;
    BSP_SPI_Init(&spi_os);
}
```

```
// BSP_SPI.c
static bsp_os_t *os;
void BSP_SPI_Init(bsp_os_t *spi_os) {
    os = spi_os;
}

void BSP_SPI_Write(uint8_t txBuf, uint32_t txLen) {
    for(int32_t i = 0; i < txLen; i++) {
        SPI1->DR = txBuf[i];
        os->pend();
    }
}

void SPI3_IRQHandler(void) {
    // Acknowledge transfer complete
    os->post();
}
```

SijWoo:
@JimothyGreene may be useful(?)

PR #214, opened 2020-09-26
Adds BSP_OS.h

SijWoo:
Read [#213](https://github.com/lhr-solar/BPS/issues/213) for more details

### DIFF
--- a/BSP/Inc/BSP_OS.h
+++ b/BSP/Inc/BSP_OS.h
@@ -1,0 +1,9 @@
+#ifndef __BSP_OS_H
+#define __BSP_OS_H
+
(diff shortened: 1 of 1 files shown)


manthanand:
Honestly, I didn't get what we were trying to do until after I left and looked over the sample code you wrote, would it make sense to pass in a parameter whenever we compile the code that selects the bare-metal or RTOS? We would still require this header file, but there would just be an if statement in the code that would load the struct for each thread with the correct functions. I just think that it would be better if we try to change the code as little as possible when we try to make it work with both systems.

ClarkPoon (approved):

PR #215, opened 2020-09-29
Voltage Mutex Initialization

sugamarora99:


Apps/Src/Voltage.c
```diff
@@ -50,7 +62,8 @@ ErrorStatus Voltage_Init(cell_asic *boards){
  */
 ErrorStatus Voltage_UpdateMeasurements(void){
 	int8_t error = 0;
-	
+	CPU_TS ts;
+	OSMutexPend(Voltage_Mutex, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
```
ClarkPoon, line 66: Can you move this to right before the function starts modifying `VoltageVal` (right before the `for` loop)? We don't want to take the mutex for longer than we need to.

Apps/Src/Voltage.c
```diff
@@ -161,11 +174,12 @@ uint32_t Voltage_GetOpenWire(void){
  * @return voltage of module at specified index
  */
 uint16_t Voltage_GetModuleMillivoltage(uint8_t moduleIdx){
-    // These if statements prevents a hardfault.
+	CPU_TS ts;
+	OSMutexPend(Voltage_Mutex, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
```
ClarkPoon, line 178: Can you move this to right before you access the `VoltageVal` array? If you pend the mutex here, there is  chance that you return from this function without posting the mutex.

ClarkPoon (changes requested):

ClarkPoon (approved):

ClarkPoon (approved):

PR #219, opened 2020-10-07
I2C Sem4

manthanand:
This branch is now done


manthanand:
Since we have only the EEPROM in our I2C, I removed the parameter in the functions that pass in the address of the EEPROM since it will always be the same. The reason it can't build is because I changed the header file.

ClarkPoon:
@manthanand can you fix the build errors?

manthanand:
The reason for the build errors is because the ifdef's in the file are not defined, so all these static functions are defined at the top of the file but there are no functions underneath. This could be fixed by adding a #define at the top of the file or compiling with the command you told me about. I thought we didn't want #defines at the top of the file so I removed them.

ClarkPoon:
> The reason for the build errors is because the ifdef's in the file are not defined, so all these static functions are defined at the top of the file but there are no functions underneath. This could be fixed by adding a #define at the top of the file or compiling with the command you told me about. I thought we didn't want #defines at the top of the file so I removed them.

Ok. I think we should either add in a `#define RTOS` or add the compiler option to define it to the makefile. What do you think @manthanand @chaseblock @SijWoo @sugamarora99

manthanand:
I think making it a compiler option would be better. Otherwise we would have to change it for every file that uses these #defines (UART, SPI so far).

ClarkPoon:
> I think making it a compiler option would be better. Otherwise we would have to change it for every file that uses these #defines (UART, SPI so far).

If we added it as a `#define` we could just have one in config.h and `#include "config.h"`  in all of the files that need it

manthanand:
Oh, but then how would we build both versions in GitHub (if you want to make sure both versions work like we do for simulator and STM)?

ClarkPoon:
> Oh, but then how would we build both versions in GitHub (if you want to make sure both versions work like we do for simulator and STM)?

To build both versions in GitHub we will probably have to edit our testing script and pass in RTOS and BAREMETAL as compiler options

BSP/Inc/BSP_I2C.h
```diff
@@ -14,22 +15,22 @@ void BSP_I2C_Init(void);
 
 /**
  * @brief   Transmits data onto the I2C bus.
- * @param   deviceAddr : the device/IC that the data is intended for.
+ * @param   devAddr : the device address to write to
```
ClarkPoon, line 18: Why did we change the parameter name from `deviceAddr` to `devAddr` the first one seems better to me.

sugamarora99 (dismissed):

ClarkPoon (approved):

ClarkPoon (approved):

PR #220, opened 2020-10-10
Voltage Open Wire Functions Fix

sugamarora99:
… to parse through the open wire variable for each minion.

### DIFF
--- a/Apps/Src/Voltage.c
+++ b/Apps/Src/Voltage.c
@@ -10,6 +10,7 @@
 
 static cell_asic *Minions;
 static uint16_t VoltageVal[NUM_BATTERY_MODULES]; //Voltage values gathered
+static uint32_t open_wires[TOTAL_PINS];
 /** LTC ADC measures with resolution of 4 decimal places, 
  * But we standardized to have 3 decimal places to work with
  * millivolts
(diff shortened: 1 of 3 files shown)


Apps/Src/Voltage.c
```diff
@@ -99,7 +99,14 @@ SafetyStatus Voltage_CheckStatus(void){
  */
 SafetyStatus *Voltage_GetModulesInDanger(void){
 	static SafetyStatus checks[NUM_BATTERY_MODULES];
-	uint32_t open_wires = Voltage_GetOpenWire();
+	Voltage_GetOpenWire();
+	uint32_t open_wires = 0;
```
ClarkPoon, line 103: Can you change this to a `static` global or a `static` global array, similar to `VoltageVal`? This way we can have a buffer to store the most recent open wire measurements

ClarkPoon (changes requested):

ClarkPoon:
@SijWoo can you review this when you get a chance?

Apps/Src/Voltage.c
```diff
@@ -99,7 +99,14 @@ SafetyStatus Voltage_CheckStatus(void){
...
+
+	//put all the bits from each minion's system_open_wire variable into one variable
+	for(int k = 0; k < NUM_MINIONS; k++){
+		open_wires += Minions[k].system_open_wire & 0x0FF;	//there are at most 8 modules per IC
```
SijWoo, line 107: Bit 0 is pin C0 which is GND. Bit 1 would be the wire between modules 1 and 2, bit 2 would be the wire between modules 2 and 3, and so on (battery modules 1-indexed).
```suggestion
		open_wires += Minions[k].system_open_wire & 0x01FF;	//there are at most 8 modules per IC 
```

Apps/Src/Voltage.c
```diff
@@ -99,7 +99,14 @@ SafetyStatus Voltage_CheckStatus(void){
...
 SafetyStatus *Voltage_GetModulesInDanger(void){
 	static SafetyStatus checks[NUM_BATTERY_MODULES];
-	uint32_t open_wires = Voltage_GetOpenWire();
+	Voltage_GetOpenWire();
+	uint32_t open_wires = 0;
+
+	//put all the bits from each minion's system_open_wire variable into one variable
+	for(int k = 0; k < NUM_MINIONS; k++){
+		open_wires += Minions[k].system_open_wire & 0x0FF;	//there are at most 8 modules per IC 
+		open_wires = open_wires << 8;
```
SijWoo, line 108: Since it's not a nice power of 2 number (should be 9), may be better to just have an array instead of an integer bitmap.

Apps/Src/Voltage.c
```diff
@@ -98,15 +99,30 @@ SafetyStatus Voltage_CheckStatus(void){
  * @return pointer to index of modules that are in danger
  */
 SafetyStatus *Voltage_GetModulesInDanger(void){
-	static SafetyStatus checks[NUM_BATTERY_MODULES];
-	uint32_t open_wires = Voltage_GetOpenWire();
+	static SafetyStatus checks[TOTAL_PINS];
```
SijWoo, line 102: To better differentiate between boards and their pins, how does the following sound?
```suggestion
	static SafetyStatus checks[NUM_MINIONS][NUM_PINS_PER_IC];
```

sugamarora99: Yeah that would be a lot more clear

Apps/Src/Voltage.c
```diff
@@ -98,15 +99,30 @@ SafetyStatus Voltage_CheckStatus(void){
...
+	Voltage_GetOpenWire();
+	uint32_t wires;
+	uint32_t openWireIdx = 0;
+	//put all the bits from each minion's system_open_wire variable into one variable
+	for(int k = 0; k < NUM_MINIONS; k++){
+		wires = (Minions[k].system_open_wire & 0x1FF);	//there are at most 8 modules per IC, bit 0 is GND
+		for(int l = 0; l < NUM_PINS_PER_IC; l++){
```
ClarkPoon, line 109: Can you use a different variable name than `l`? It looks a lot like the number `1` in some fonts

ClarkPoon (changes requested):
Just some style-related changes

ClarkPoon (approved):

PR #226, opened 2020-10-12
CAN Interrupts and OS Integration

chaseblock:
Problem
=======
The CAN driver and BSP code was previously using polling-like
techniques to send and receive messages. This will not integrate
well with the move to uC/OS.

Solution
========
The CAN drivers are being moved to an interrupt-driven system,
whereby the receiving messages come via interrupt for later collection,
and the transmitted messages utilize interrupts and locks to ensure
thread safety.

There are now two methods for transmitting messages. These are:
* CANbus_Send, which preserves closely its old functionality for backwards compatibility.
* CANbus_BlockAndSend, which puts the calling thread to sleep until the message can be sent.

If the first of the above messages is called, the caller is expected
to check for a proper transmission. If a guarenteed transmission is
required, the second function should be called.

To receive messages, there are likewise two functions: one nonblocking and
one blocking (CANbus_Receive & CANbus_WaitToReceive). Use the first one
only if you do not want to block and do not plan to poll.

Notes
=====
No testing has been done on these yet, and there has been no work on the
simulator portions.


chaseblock:
Re-opening, since I previously opened a non-draft by mistake.

The reason this doesn't build the tests is because the simulator portion has not been updated.

Drivers/Src/CANbus.c
```diff
@@ -1,75 +1,264 @@
...
-    uint8_t txdata[5];
+// Static method, call CANbus_Send or CANbus_BlockAndSend instead
+static int CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
+	uint8_t txdata[5];
+	uint8_t data_length = 0;
 	
+	OS_ERR err;
+	CPU_TS ts;
+
+	// TODO: is it really best to keep the list of
```
ClarkPoon, line 83: I'm open to better ideas if you have any. If not, I'm fine keeping it this way.

Drivers/Src/CANbus.c
```diff
@@ -1,75 +1,264 @@
...
-/**
- * @brief   Transmits data onto the CANbus
- * @param   id : CAN id of the message
- * @param   payload : the data that will be sent.
- * @return  0 if data wasn't sent, otherwise it was sent.
- */
-int CANbus_Send(CANId_t id, CANPayload_t payload) {
-    uint8_t txdata[5];
+// Static method, call CANbus_Send or CANbus_BlockAndSend instead
+static int CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
```
ClarkPoon, line 76: Why does this return an `int` instead of `ErrorStatus` or `uint8_t` (return type of `BSP_CAN_Write()`. Also, the same thing for all of the following functions that return `int` as an error code

ClarkPoon (commented):
Some coding style suggestions

sugamarora99 (approved):

PR #227, opened 2020-10-13
Rtos safetycheck and fault sem4s

SijWoo:
Adds Safety Check and Fault semaphores. I tested with abstracted Voltage, Temperature, and Amperes functions.

_Notes:_
Debugger for stm32f413 needs to be tested with rtos code. For some reason I get errors while in gdb that I haven't seen before but I believe code is still running on the MCU. Will make a separate issue ticket.


Apps/Inc/Tasks.h
```diff
@@ -57,6 +57,41 @@ void Task_BLE(void *p_arg);
...
+extern CPU_STK CriticalState_Stk[TASK_CANBUS_CONSUMER_STACK_SIZE];
+extern CPU_STK PetWDog_Stk[TASK_PETWDOG_STACK_SIZE];
+extern CPU_STK VoltTempMonitor_Stk[TASK_VOLT_TEMP_MONITOR_STACK_SIZE];
+extern CPU_STK AmperesMonitor_Stk[TASK_AMPERES_MONITOR_STACK_SIZE];
+extern CPU_STK LogInfo_Stk[TASK_LOG_INFO_STACK_SIZE];
+extern CPU_STK CANBusConsumer_Stk[TASK_CANBUS_CONSUMER_STACK_SIZE];
+extern CPU_STK BatteryBalance_Stk[TASK_BATTERY_BALANCE_STACK_SIZE];
+extern CPU_STK CLI_Stk[TASK_CLI_STACK_SIZE];
+extern CPU_STK BLE_Stk[TASK_BLE_STACK_SIZE];
+extern CPU_STK Idle_Stk[TASK_IDLE_STACK_SIZE];
```
chaseblock, line 88: Where are these definitions used? Do they really need to be declared extern here?

Apps/Src/Current.c
```diff
@@ -107,10 +107,27 @@ void Task_AmperesMonitor(void *p_arg) {
 
     OS_ERR err;
 
+	bool isfirst_amperes_check = false;
```
chaseblock, line 110: Maybe it would be better to rename this to something like `amperesHasBeenChecked`? Right now, the name and what it's being used for seem a bit backwards to me.

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -1,22 +1,78 @@
...
 void Task_VoltTempMonitor(void *p_arg) {
     (void)p_arg;
 
     OS_ERR err;
 
+    // SafetyCheck_Sem4 must only be signaled once per parameter at system boot up.
+    // These flags indicate was signaled for that parameter
+    bool isfirst_voltage_check = false;
+    bool isfirst_openwire_Check = false;
+    bool isfirst_temperature_check = false;
```
chaseblock, line 16: Same comment as on the Amperes version.

Apps/Src/main.c
```diff
@@ -14,6 +14,10 @@ int main() {
 	
 	OS_ERR err;
 
+	OSInit(&err);
+
+	// ASSERT err
```
chaseblock, line 19: Are we missing the assert function?

Apps/Src/Task_VoltTempMonitor.c
ClarkPoon, line 16: See comment on `isfirst_amperes_check`

ClarkPoon (changes requested):

Apps/Inc/Tasks.h
SijWoo: No they don't have to be here, but all the task function declarations and other task information were already in Tasks.h so I tried to stay consistent with what is already present in the rtos branch.

Apps/Src/Current.c
SijWoo: Sounds gud

Apps/Src/main.c
SijWoo: The assert wasn't merged yet when I did this but now I see the rtos branch now has it. I'll update it

Apps/Src/Current.c
```diff
@@ -107,10 +107,27 @@ void Task_AmperesMonitor(void *p_arg) {
...
+
     while(1) {
         // BLOCKING =====================
         // Update Amperes Measurements
+		Current_UpdateMeasurements();
+
         // Check if amperes is NOT safe:
+		SafetyStatus amperes_status = Current_CheckStatus(true);
```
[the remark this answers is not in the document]
SijWoo: This was copied from the bare metal code. I'll update this but this is also present in master branch

Apps/Src/Current.c
```diff
@@ -107,10 +107,27 @@ void Task_AmperesMonitor(void *p_arg) {
...
+            OSSemPost(&Fault_Sem4,
+                        OS_OPT_POST_1,
+                        &err);
+        } else if((amperes_status == SAFE) && (!isfirst_amperes_check)) {
+            // Signal to turn on contactor but only signal once
+            OSSemPost(&SafetyCheck_Sem4,
+                        OS_OPT_POST_1,
+                        &err);
+            // assert
+            isfirst_amperes_check = true;
```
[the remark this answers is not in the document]
SijWoo: oh yee lol, Chase's suggestion will fix this confusion

Apps/Src/Current.c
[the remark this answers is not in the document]
SijWoo: Hm actually, the Current_CheckStatus implementation may need to modified in another ticket.

Apps/Src/Current.c
```diff
@@ -108,10 +107,27 @@ void Task_AmperesMonitor(void *p_arg) {
...
 
+	bool amperesHasBeenChecked = false;
+
     while(1) {
         // BLOCKING =====================
         // Update Amperes Measurements
+		Current_UpdateMeasurements();
+
         // Check if amperes is NOT safe:
+		SafetyStatus amperesStatus = Current_CheckStatus(true);
```
ClarkPoon, line 118: If we are using Voltage.c to catch the case where we are discharging during an under-voltage and for this to check the full range of currents, we should pass `false` to this instead of `true`.

ClarkPoon (changes requested):

ClarkPoon (approved):

ClarkPoon (approved):

PR #229, opened 2020-10-17
Reseting timer if all tasks are set

harshigorla:


Apps/Src/Task_PetWDog.c
```diff
@@ -14,7 +14,10 @@ void Task_PetWDog(void *p_arg) {
     // If WDog_BitMap is all set:
     //      Reset watchdog timer
 
-
+     if (WDog_BitMap == 7) {
```
ClarkPoon, line 17: Can you add a comment on why we are checking if it is equal to 7?

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
ClarkPoon, line 1: Can you remove the .vs directory from this pull request?

ClarkPoon (changes requested):

Apps/Src/Task_PetWDog.c
harshigorla: I added the comment and committed and pushed it but I can't see the pull request option this time. The file is under commits tho.

ClarkPoon: Ok, this part looks good. Once you open a pull request, it will track your branch, so you don't have to reopen it after each commit you make

ClarkPoon (approved):

PR #232, opened 2020-10-17
Critical task

manthanand:


Apps/Src/Task_CriticalState.c
```diff
@@ -1,26 +1,28 @@
...
+    while(OSSemPend(&SafetyCheck_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err) != 4);
     // Turn Contactor On
-    
+    BSP_Contactor_On();
     // Push All Clear message to CAN Q
-
+    CANbus_Send(ALL_CLEAR, Can_MSG);
     // Push Contactor State message to CAN Q
+    CANbus_Send(CONTACTOR_STATE, Can_MSG);
 }
```
SijWoo, line 28: Add the OS task delete function at the end just in case so the RTOS don't crash

Apps/Src/Task_CriticalState.c
```diff
@@ -1,26 +1,28 @@
...
-
+    while(OSSemPend(&SafetyCheck_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err) != 4);
     // Turn Contactor On
-    
+    BSP_Contactor_On();
     // Push All Clear message to CAN Q
-
+    CANbus_Send(ALL_CLEAR, Can_MSG);
     // Push Contactor State message to CAN Q
+    CANbus_Send(CONTACTOR_STATE, Can_MSG);
```
SijWoo, line 27: With the way the CANBusConsumer task is planned out, the CAN messages should be pushed onto the CAN Bus queue which the CANBusCOnsumer task will process and call CANbus_Send. The CANBusConsumer task was added because multiple threads will be calling the CAN bus functions so the the queue will let the threads send CAN messages without blocking itself

SijWoo: Not sure if the CANBus Consumer task has been implemented or not

manthanand: Should the pushing of the message into the queue be implemented in canbus.h or somewhere else?

ClarkPoon (approved):

PR #233, opened 2020-10-17
 threadInTaskInit

alekxoo:


Apps/Src/Task_Init.c
```diff
@@ -23,4 +22,172 @@ void Task_Init(void *p_arg) {
...
+				(void *)0,				// Task function args
+				TASK_IDLE_PRIO,			// Priority
+				Init_Stk,				// Stack
+				WATERMARK_STACK_LIMIT,	// Watermark limit for debugging
+				DEFAULT_STACK_SIZE,		// Stack size
+				0,						// Queue size (not needed)
+				10,						// Time quanta (time slice) 10 ticks
+				(void *)0,				// Extension pointer (not needed)
+				OS_OPT_TASK_STK_CHK | OS_OPT_TASK_SAVE_FP,	// Options
+				&err);					// return err code}
```
manthanand, line 192: This thread should destroy itself after it is done.

Apps/Src/Task_Init.c
```diff
@@ -4,9 +4,8 @@
 void Task_Init(void *p_arg) {
 
     OS_ERR err;
-
-    OSSemCreate(&SafetyCheck_Sem4,
-                "Safety Check Semaphore",
+    OSSemCreate(&Fault_Sem4,
```
manthanand, line 7: Why is this not safety check sem4?

Apps/Src/Task_Init.c
```diff
@@ -23,4 +22,172 @@ void Task_Init(void *p_arg) {
     assertOSError(err);
 
     // TODO: Initialize threads! Look at main.c for reference
+        //0
+    	OSTaskCreate(&Init_TCB,				// TCB
```
manthanand, line 26: Why are these all Init_TCB, they should all be TCB's for their own thread. These are all listed in Tasks.c

Apps/Src/Task_Init.c
```diff
@@ -23,4 +22,172 @@ void Task_Init(void *p_arg) {
...
+				"Initialize System",	// Task Name (String)
+				Task_Init,				// Task function pointer
+				(void *)0,				// Task function args
+				TASK_INIT_PRIO,			// Priority
+				Init_Stk,				// Stack
```
manthanand, line 31: They also all have their own stack sizes.

ClarkPoon (changes requested):

Apps/Src/Task_Init.c
```diff
@@ -23,4 +23,160 @@ void Task_Init(void *p_arg) {
     assertOSError(err);
 
     // TODO: Initialize threads! Look at main.c for reference
```
ClarkPoon, line 25: We should remove this comment, since it has already been done

ClarkPoon (commented):
Do you know how to get rid of the changes to the submodules, or should I do that?

manthanand:
I think we need those changes to the submodules in order for the code to work

ClarkPoon:
> I think we need those changes to the submodules in order for the code to work

Yeah, you're right

PR #234, opened 2020-10-17
Wrote SPI1 Handler that posts the MinionsIO_Sem4. 

sugamarora99:
…_Init() and a new BSP_SPI_Write() for RTOS code.


Drivers/Src/LTC6811.c
```diff
@@ -72,12 +72,34 @@ Copyright 2017 Linear Technology Corp. (LTC)
...
+// RTOS Setup
+void LTC6811_Pend(void) {
+    CPU_TS ts;
+    OS_ERR err;
+    OSSemPend(&MinionsIO_Sem4,
+                        0,
+                        OS_OPT_PEND_BLOCKING,
+                        &ts,
+                        &err);
+}
```
manthanand, line 95: Inside these functions, you should also call the assertOS function(not sure if that is the correct name but it should be in tasks.c.

Drivers/Src/LTC6811.c
```diff
@@ -72,12 +72,34 @@ Copyright 2017 Linear Technology Corp. (LTC)
...
 #include "Tasks.h"
 
 /*********************************************************/
 /*** Code that was added by UTSVT. ***/
 /*********************************************************/
 OS_MUTEX MinionsASIC_Mutex;
+bsp_os_t spi_os;
+OS_SEM MinionsIO_Sem4;
+
+#define RTOS
```
manthanand, line 85: Why do you #define RTOS here?

BSP/Inc/BSP_SPI.h
```diff
@@ -16,6 +17,23 @@
...
+
+/**
+ * @brief   Initializes the SPI port connected to the LTC6820.
+ *          This port communicates with the LTC6811 voltage and temperature
+ *          monitoring IC. The LTC6820 converts the SPI pins to 2-wire isolated SPI.
+ *          Look at analog devices website and LTC6811's or LTC6820's datasheets.
+ * @param   None
+ * @return  None
+ */
+void BSP_SPI_Init(bsp_os_t *spi_os);
```
manthanand, line 35: Instead of having this ifdef, just make it so that way the function always accepts an input of that type. When we run the baremetal version of the code, we can just load the pend and post functions with empty code. It's neater that way.

manthanand:
Were you supposed to implement the SPI IRQ in this pull request?

Drivers/Src/LTC6811.c
```diff
@@ -92,13 +114,17 @@ void LTC6811_Init(cell_asic *battMod){
     mutexExists = true;
   }
 
-	BSP_SPI_Init();				// Initialize SPI1 for voltage board	
-	
+  spi_os.pend = LTC6811_Pend;
+  spi_os.post = LTC6811_Post;
+	BSP_SPI_Init(&spi_os);				// Initialize SPI1 for voltage board
```
manthanand, line 119: In the actual BSP_Init function, you should load these function memory addresses to something in the library. Then whenever you access Minions, you would call those functions and they would run the pend and post. Also, what is the variable that this semaphore is being used on? I don't see one in BSP_SPI.c.

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -126,3 +133,88 @@ void BSP_SPI_SetStateCS(uint8_t state) {
...
+	GPIO_InitStruct.GPIO_Speed = GPIO_Speed_50MHz;
+	GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;
+    GPIO_Init(GPIOB, &GPIO_InitStruct);
+	
+	os = spi_os;
+}
+
+void BSP_SPI_Write(uint8_t txBuf, uint32_t txLen){
+	for(int32_t i = 0; i < txLen; i++){
+		SPI_WriteRead(txBuf[i]);
```
manthanand, line 214: Shouldn't you pend before the for loop and post after the for loop?

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -126,3 +133,88 @@ void BSP_SPI_SetStateCS(uint8_t state) {
...
+	// Save the CPU registers
+	CPU_SR_ALLOC();
+
+	// Protect a critical section
+	CPU_CRITICAL_ENTER();
+
+	// make the kernel aware that the interrupt has started
+	OSIntEnter();
+	CPU_CRITICAL_EXIT();
+	os->pend();
```
manthanand, line 151: I guess we went over this in the meeting but I'm still not sure what is supposed to happen in this interrupt. All the stuff you have already done is so the interrupt can run in an RTOS. Also, you should probably post whatever you are pending.

sugamarora99: Yeah this is a typo I will fix this soon

Drivers/Src/LTC6811.c
```diff
@@ -92,13 +115,17 @@ void LTC6811_Init(cell_asic *battMod){
     mutexExists = true;
   }
```
manthanand, line 117: In here, I think you would have to have an ifdef which sets the pend and post to empty functions for baremetal. EX.
/***BAREMETAL VERSION OF CODE**/
LTC6811_Pend(){ return;}
So you would have two pend and post functions and depending on which version we are compiling. That is the version that would be placed in the struct.

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -16,6 +20,8 @@ static uint8_t SPI_WriteRead(uint8_t txData){
 	return SPI1->DR & 0x00FF;
 }
 
+
+#ifdef BAREMETAL
```
manthanand, line 24: With the way this is set up. If we compile with RTOS, then the SPI_Init function will not be compiled.

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -126,3 +135,34 @@ void BSP_SPI_SetStateCS(uint8_t state) {
         GPIO_ResetBits(GPIOB, GPIO_Pin_6);
     }
 }
+
+/***************THE FOLLWING CODE IS FOR THE RTOS VERSION OF THE BPS*********/
+#ifdef RTOS
+
+void SPI1_IRQHandler(void){
```
manthanand, line 142: I'm not sure if this was on purpose, but if the ifdef RTOS is here, then we would not have an IRQ for SPI when using baremetal code. If we do not want an IRQ for SPI with bare metal than this will work

sugamarora99:
This branch won't compile until either BAREMETAL or RTOS is defined- this is why the checks have failed

manthanand (dismissed):

sugamarora99:
This PR also includes the SPI3 handler and AmperesIO semaphore.

PR #240, opened 2020-11-07
Faulttask

manthanand:
Set fans to full speed when in fault state


manthanand:
This branch requires the Fan Drivers branch in order to build properly.

ClarkPoon (commented):
Why does it show the submodules as modified? @SijWoo do you know what is going on here?

ClarkPoon (approved):

PR #241, opened 2020-11-07
Decreases max temperature limit to INR's datasheets recommended relea…

SijWoo:
Decreases max **discharge** temperature limit to 55.00C. The datasheet recommends release during discharge of <60.00C.

Decreases max **charge** temperature limit to 40.00C. The datasheet recommends release during charge of <45.00C.


manthanand:
Should we also include the max temperature differential between modules?

ClarkPoon (approved):

PR #244, opened 2020-11-14
Changed GetModulesInDanger Function in Voltage.c

sugamarora99:
Voltage_GetModulesInDanger() now returns a struct that has two SafetyStatus arrays: one for the modules and one for the wires in our system.

### DIFF
--- a/Apps/Inc/Voltage.h
+++ b/Apps/Inc/Voltage.h
@@ -17,7 +17,7 @@
 typedef struct system_danger{
     SafetyStatus wire_checks[TOTAL_VOLT_WIRES];
     SafetyStatus module_checks[NUM_BATTERY_MODULES];
-} Safety_Info;
+} Voltage_Safety;
 
 /** Voltage_Init
  * Initializes all device drivers including LTC6811 and GPIO to begin Voltage Monitoring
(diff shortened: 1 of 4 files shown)


Apps/Inc/Voltage.h
```diff
@@ -14,6 +14,11 @@
 #include "common.h"
 #include "os.h"
 
+typedef struct system_danger{
+    SafetyStatus wire_checks[TOTAL_VOLT_WIRES];
+    SafetyStatus module_checks[NUM_BATTERY_MODULES];
+} Safety_Info;
```
SijWoo, line 20: I would make the struct more specific to Voltage.c since Safety_Info could be applied to Temperature.c and Amperes.c. Maybe VoltageSafetyChecks or something that has "voltage" in the same

Apps/Src/Voltage.c
```diff
@@ -162,16 +162,16 @@ SafetyStatus *Voltage_GetModulesInDanger(void){
 		if(i < NUM_BATTERY_MODULES){
 			// Check if battery is in range of voltage limit
 			if(Voltage_GetModuleMillivoltage(i) > MAX_VOLTAGE_LIMIT * MILLI_SCALING_FACTOR || Voltage_GetModuleMillivoltage(i) < MIN_VOLTAGE_LIMIT * MILLI_SCALING_FACTOR) {
```
SijWoo, line 164: This could be done later but we should separate this so we can specify UNDERVOLTAGE or OVERVOLTAGE dangers

ClarkPoon (changes requested):
We need to fix the submodules before I can merge this

ClarkPoon (approved):
Just realized the submodule changes are necessary

PR #246, opened 2020-11-14
Added undervoltage cutoff detection for charging

manthanand:


Apps/Src/Voltage.c
```diff
@@ -126,13 +127,11 @@ SafetyStatus Voltage_CheckStatus(void){
...
 		// VOLTAGE_LIMITS are in floating point. The LTC6811 sends the voltage data
 		// as unsigned 16-bit fixed point integers with a resolution of 0.00001
-		if(voltage > MAX_VOLTAGE_LIMIT * MILLI_SCALING_FACTOR){
-			return OVERVOLTAGE;
-		}
-		
-		else if(voltage < MIN_VOLTAGE_LIMIT * MILLI_SCALING_FACTOR){
-			return UNDERVOLTAGE;
+		if(voltage > MAX_VOLTAGE_LIMIT * MILLI_SCALING_FACTOR) return OVERVOLTAGE;
+		if (Current_IsCharging){
```
ClarkPoon, line 131: Shouldn't this be `if (Current_IsCharging()){`

ClarkPoon (changes requested):
Can you get rid of the changes to the submodules, or should I do that?

manthanand:
I don't think it'll let me do it so you probably should.

ClarkPoon (approved):
After a second look, the changes to the submodules in here are correct

PR #247, opened 2020-11-14
Log info edits

harshigorla:
Sends the percentage of charge left on the battery pack to EEPROM every 3 seconds


Apps/Src/Task_PetWDog.c
```diff
@@ -14,7 +14,10 @@ void Task_PetWDog(void *p_arg) {
     // If WDog_BitMap is all set:
     //      Reset watchdog timer
 
-
+     if (WDog_BitMap == 7) {
```
ClarkPoon, line 17: Not sure how this change got in here, since I thought we already merged this into the rtos branch

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
ClarkPoon, line 1: Can you not add the .vs files or binary files (.suo, .db, .ipch, .sqlite) in the future?

ClarkPoon (changes requested):

ClarkPoon (approved):

PR #253, opened 2020-11-21
Fixed some compile warnings

ClarkPoon:


Apps/Src/Task_LogInfo.c
```diff
@@ -12,7 +13,7 @@ void Task_LogInfo(void *p_arg) {
     EEPROM_LogData(DATA_SOC, Charge_GetPercent());
 
     //1 tick every 10 millisconds, so 300 ticks for a delay of 3000ms --> 3s
-    //OS_OPT_TIME_DLY specifies relative delay, so the delay is started from the �current time + dly�.
+    //OS_OPT_TIME_DLY specifies relative delay, so the delay is started from the �current time + dly�.
```
SijWoo, line 16: Is it just me or have these quotation marks been changed to diamonds with questions marks in them?

Apps/Src/Task_PetWDog.c
```diff
@@ -1,6 +1,7 @@
 #include "Tasks.h"
 #include "os.h"
 #include <stdlib.h>
```
SijWoo, line 3: Why was stdlib included? This slipped through from another PR :P

Apps/Src/Task_LogInfo.c
ClarkPoon: I'll look into this

SijWoo (approved):
noice

PR #258, opened 2020-11-22
Can consumer

manthanand:
CanConsumer task finished by Sugam. I added code to VoltTempMonitor and Current task to send data across CAN. It will not build until Chase's branch is merged.


Apps/Src/Current.c
```diff
@@ -184,11 +185,22 @@ void Task_AmperesMonitor(void *p_arg) {
 
 	bool amperesHasBeenChecked = false;
 
+	CANData_t CanData;
+    CanData.b = 0;
+    CanData.h = 0;
+    CanData.w = 0;
```
ClarkPoon, line 191: You only have to set one of these to 0, since this is a union

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -19,11 +20,24 @@ void Task_VoltTempMonitor(void *p_arg) {
     bool openWireHasBeenChecked = false;
     bool temperatureHasBeenChecked = false;
 
+    CANData_t CanData;
+    CanData.b = 0;
+    CanData.h = 0;
+    CanData.w = 0;
```
ClarkPoon, line 26: In a union, all of the fields share the same memory, so you only have to set one of the fields

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -19,11 +20,24 @@ void Task_VoltTempMonitor(void *p_arg) {
...
+    CANMSG_t CanMsg;
     while(1) {
         // BLOCKING =====================
         // Update Voltage Measurements
         Voltage_UpdateMeasurements();
-        
+        CanMsg.id = VOLT_DATA;
+        for (int i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module voltage data
+            CanPayload.idx = i;
+            CanData.f = Voltage_GetModuleMillivoltage(i);
```
ClarkPoon, line 36: This function returns a `uint16_t` not a `float`

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -65,7 +79,14 @@ void Task_VoltTempMonitor(void *p_arg) {
         // BLOCKING =====================
         // Update Temperature Measurements
         Temperature_UpdateAllMeasurements();
-        
+        CanMsg.id = TEMP_DATA;
+        for (int i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module temp data
+            CanPayload.idx = i;
+            CanData.f = Temperature_GetModuleTemperature(i);
```
ClarkPoon, line 85: This function returns an `int32_t`

Apps/Src/Current.c
```diff
@@ -184,11 +185,22 @@ void Task_AmperesMonitor(void *p_arg) {
...
+    CANPayload_t CanPayload;
+    CANMSG_t CanMsg;
+
     while(1) {
         // BLOCKING =====================
         // Update Amperes Measurements
 		Current_UpdateMeasurements();
-
+		CanData.f = (float)Amps_GetReading();
```
ClarkPoon, line 199: This function returns an `int32_t` that represents a fixed point number, so you can't cast it like this

ClarkPoon (changes requested):

Apps/Src/Current.c
manthanand: So would I just return the value of Amps_GetReading()? Or would I have to assign that to an integer and then typecast that variable. Do I have to do the same for all of the other CANData's?

ClarkPoon: Maybe cast it to a float, then divide it by 1000 to get Volts? It doesn't make sense to use a float if we are not using the fraction part of it

manthanand:
I'm confused by why we are sending floating-point data across the CAN when we could just send the data in it's original form. Is it necessary for it to be in floating point or can we store it somewhere else since it will be a union anyways? Also, in the comments where the CANData is defined, it says only one of the places should be given data. That is why I made all the other values 0 (instead of just one).

Apps/Src/Current.c
ClarkPoon: Also, the Amps pull request has been merged into rtos, so you can merge rtos into this and get it to build

ClarkPoon:
> I'm confused by why we are sending floating-point data across the CAN when we could just send the data in it's original form. Is it necessary for it to be in floating point or can we store it somewhere else since it will be a union anyways? Also, in the comments where the CANData is defined, it says only one of the places should be given data. That is why I made all the other values 0 (instead of just one).

I think we are using floats so we can send volts with decimals instead of millivolts. If we just want to send data in units of millivolts, then yes it makes more sense to send an integer

Apps/Src/Task_VoltTempMonitor.c
manthanand: In that case, I will only set the one that has data in it and I will remove all the 0's.

Apps/Src/Task_VoltTempMonitor.c
manthanand: How should I convert this to a float? I'm not sure how the data is being read (are all 32 bits meant to be for 2 digit celsius numbers)?

ClarkPoon (approved):

PR #263, opened 2020-12-06
Warnings

chaseblock:


BSP/Simulator/Src/BSP_UART.c
```diff
@@ -209,9 +209,15 @@ static bool RxFifo_Peek(uint8_t *data, UART_Port usart) {
...
+
+    // In event of an error, assume it is full
+    return true;
 }
 
 static bool RxFifo_IsEmpty(UART_Port usart) {
     if (usart == UART_USB) return rxGet3 == rxPut3;
     if (usart == UART_BLE) return rxGet2 == rxPut2;
+
+    // In event of an error, assume it is not empty
```
ClarkPoon, line 221: Why are we choosing these assumptions if we have an error? It seems like we would want to return 'true' for both error cases, since `RxFifo_Get()` and `RxFifo_Put()` do nothing if we return `true`

ClarkPoon (approved):

PR #265, opened 2020-12-12
Battery Balancing Task

sugamarora99:


Apps/Src/Task_BatteryBalance.c
```diff
@@ -1,15 +1,17 @@
...
 void Task_BatteryBalance(void *p_arg) {
     (void)p_arg;
 
     OS_ERR err;
 
-    while(1) {
-        // Todo: Battery balancing
-
+    while(1){
+        Balancing_Balance(Minions);
```
ClarkPoon, line 14: If you access the Minions array, you need to use the `MinionsASIC_Mutex` (see Voltage.c). I think it would probably be best to add the mutex calls to the functions in Balancing.c that access this array, since if you add it on the task level, this will stop the voltage and temperature thread from running most of the time

ClarkPoon (changes requested):

ClarkPoon (approved):

ClarkPoon (approved):

PR #270, opened 2020-12-14
Cdr changes

ClarkPoon:
I made some updates based on feedback from CDR. I also fixed some bugs and cleaned up some code that I saw while I was working on this. The main changes I made are:
- removed dashboard notify task
- changed some shutoff limits
- removed some commented out code
- moved the voltage, temperature, and current CAN message creation to after determining if the battery is safe


manthanand:
I just wanna make sure that our plan was to decrease both temperature thresholds and not just one. I was under the impression that we were trying to have it at 2.55 for charging and 2.9 for discharging so we would have a larger gap between the two. If we want them both to have the same gap but just at a lower threshold then I think that is fine too.

ClarkPoon:
> I just wanna make sure that our plan was to decrease both temperature thresholds and not just one. I was under the impression that we were trying to have it at 2.55 for charging and 2.9 for discharging so we would have a larger gap between the two. If we want them both to have the same gap but just at a lower threshold then I think that is fine too.

The plan was to decrease both of the voltage thresholds, so we can use more of our battery's charge. The gap is slightly smaller, but I think it should still be fine, since the open circuit voltage will be greater than the voltage when we are powering the motor, so we shouldn't see a drop of 0.15V after tripping the BPS

chaseblock (approved):

sugamarora99 (approved):

PR #274, opened 2020-12-21
Rtos issues

manthanand:


manthanand:
I went into the rtos branch and the same build error occurs there

Apps/Src/Task_CriticalState.c
```diff
@@ -6,6 +7,10 @@
 #include "BSP_UART.h"
 #include "Voltage.h"
 
+/*
+This flag is used to override the undervoltage fault condition of the BPS when we are starting
+the car(if we need to charge). If it is true, then we should be in charging state only.
+*/
 bool AdminOverride = false;
```
ClarkPoon, line 14: I think this violates regulation 5.2.E.9: "Firmware or settings will be rendered static and un-modifiable after inspection". It should be safe to remove this and everything that uses it, since we have two separate under voltage cutoffs for charging and discharging. I forgot to remove this when I removed the dashboard override task

Apps/Src/Voltage.c
```diff
@@ -161,19 +160,18 @@ Voltage_Safety Voltage_GetModulesInDanger(void){
 		if(i < NUM_BATTERY_MODULES){
 			// Check if battery is in range of voltage limit
 			if(Voltage_GetModuleMillivoltage(i) > MAX_VOLTAGE_LIMIT * MILLI_SCALING_FACTOR) {
-				system.module_checks[i] = OVERVOLTAGE;
+				(*system).module_checks[i] = OVERVOLTAGE;
```
ClarkPoon, line 163: An equivalent line of code that I think looks cleaner is
```C
system->module_checks[i] = OVERVOLTAGE;
```

ClarkPoon (changes requested):

ClarkPoon (dismissed):

ClarkPoon (approved):

PR #277, opened 2020-12-25
Replaced tabs with spaces

ClarkPoon:
My script replaced all tabs with four spaces in all .c and .h files. It says that it succeeded for all files except:
1. BPS/BSP/STM32F413/STM32F4xx_StdPeriph_Driver/Src/stm32f4xx_qspi.c
2. BPS/CMSIS/Device/ST/STM32F4xx/Source/Templates/system_stm32f4xx.c
3. BPS/RTOS/uCOS-III-STM32F4/uCOS-III/Ports/ARM-Cortex-M4/Generic/IAR/os_cpu_c.c

I think the failures were caused by non-unicode characters in these files, which broke Python's file.read() method. The error thrown was:

`UnicodeDecodeError: 'utf8' codec can't decode byte <something I forgot to write down> in position <something I forgot to write down>: invalid start byte`

I didn't feel like fixing this error because we will probably not edit these files. I used the following code to do this:

```Python
import os

log = []

def clean(file):
    with open(file, 'r+') as f:
        try:
            in_buffer = f.read()
            out_buffer = ''
            for c in in_buffer:
                if c == '\t':
                    out_buffer += ('    ')
                else:
                    out_buffer += c
            f.write(out_buffer)
        except:
            log.append('error reading ' + file)


def recurse_and_clean(directory):
    members = os.listdir(directory)
    for m in members:
        if (os.path.isdir(directory + '/' + m)):
            print('recursing into ' + m)
            recurse_and_clean(directory + '/' + m)
        elif (m[-2:] == '.c' or m[-2:] == '.h'):
            print('cleaning ' + m)
            clean(directory + '/' + m)
            print('cleaned ' + m)

rootdir = os.getcwd()
recurse_and_clean(rootdir)

print('----------------------------------------------')
print('log:')
for error in log:
    print(error)
print('----------------------------------------------')

print('done!')
```


chaseblock (changes requested):
Just looking at the first few files, this obviously did not replace all the tabs, but rather doubled all content in the files.

ClarkPoon:
> Just looking at the first few files, this obviously did not replace all the tabs, but rather doubled all content in the files.

Yeah, I realized that this is busted a few minutes after opening the PR, but haven't gotten around to fixing it yet

chaseblock:
You're also going to have to push to the submodules separately in order for this to build.

ClarkPoon:
> You're also going to have to push to the submodules separately in order for this to build.

Should we be doing pull requests and code reviews for the submodules as well?

chaseblock:
Yeah, still good to make sure those don't break. Eventually, those should have someone from Controls also sign off, but right now I think they're only used by BPS.

ClarkPoon:
I think this is low priority enough that the changes here will be irrelevant (because of other code changes) by the time we actually get around to this

PR #294, opened 2021-01-23
EEPROM Drivers docs

ClarkPoon:
I am planning on using this as the example for the drivers docs, so feedback on the format would be helpful

### DIFF
--- a/Docs/source/Software/Drivers.rst
+++ b/Docs/source/Software/Drivers.rst
@@ -1,0 +1,19 @@
+*******
+Drivers
+*******
+
+EEPROM Driver
+=============
+
+Purpose
+    The EEPROM Driver is used to read and write the the external EEPROM on the leader board. We are using the EEPROM to store the state of charge of the battery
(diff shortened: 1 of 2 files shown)


Docs/source/Software/Drivers.rst
```diff
@@ -0,0 +1,19 @@
...
+
+EEPROM Driver
+=============
+
+Purpose
+    The EEPROM Driver is used to read and write the the external EEPROM on the leader board. We are using the EEPROM to store the state of charge of the battery
+    and the cause of any faults that cause the BPS to trip. We are using an EEPROM to store this information so it will survive a BPS reset.
+
+Usage
+    ``EEPROM_Init()`` must be called before any of the other EEPROM functions can be used. ``EEPROM_Load()`` or ``EEPROM_Reset()`` must be called before
```
manthanand, line 13: I'm not sure if we should include this in here or in hardware(or both), but we should probably write somewhere how often to read/write to the EEPROM.

sugamarora99 (approved):

manthanand (approved):

PR #302, opened 2021-01-23
Temperature App Documentation

sugamarora99:


manthanand (dismissed):

Docs/source/Software/Apps.rst
```diff
@@ -173,6 +173,22 @@ Usage
...
 
+Temperature
+===========
+
+Purpose
+    The BPS must monitor the battery pack's temperature to avoid any hazardous conditions for the car and the driver.
+    Each battery module has 2 temperature sensors.
+
+Usage
+    This file uses a global ``int32_t ModuleTemperatures[][]`` 2D array that contains each sensor's temperature.
```
ClarkPoon, line 184: I think it is important to note that you should call ``Temperature_UpdateAllMeasurements()`` to update this array. For example, it is not obvious from reading the header file that you need to call this function to get the data that ``Temperature_CheckStatus()`` uses.

ClarkPoon (changes requested):

ClarkPoon (dismissed):

ClarkPoon (approved):

PR #307, opened 2021-01-30
Pet WatchDog Task documentation

manthanand:
I left out that it yields if the bits aren't set and the watchdog faults because that should be self-explanatory.

### DIFF
--- a/Docs/source/Software/Tasks.rst
+++ b/Docs/source/Software/Tasks.rst
@@ -217,7 +217,7 @@ Yields
     It doesn't yield.
 
 Additional Considerations
-    If we add more tasks(or split up tasks such as voltage and temperature) and want to have the 
+    If we add more tasks (or split up tasks such as voltage and temperature) and want to have the 
     watchdog timer look over them, we can add more bits to the timer and just check if they are set.
 
 Idle Task
--- a/Docs/source/Software/Tasks.rst
+++ b/Docs/source/Software/Tasks.rst
@@ -211,7 +211,7 @@ Shared Resources
     BatteryBalancing tasks.
 
 Timing Requirements
-    This task is set to run every 40 milliseconds.
+    This task is set to run every 400 milliseconds.
 
 Yields
     It doesn't yield.
--- a/Docs/source/Software/Tasks.rst
+++ b/Docs/source/Software/Tasks.rst
@@ -198,7 +198,7 @@ Functionality
     This task checks the ``WDog_BitMap`` variable to see if the 3 LSB are set by the temperature,
     voltage, balancing, and current tasks. If these tasks ran and set those bits, that means that the 
     BPS is functional and the timer is reset. If those bits are not set, the timer will keep running
-    and a fault handler will be called if it reaches 0. The task can be called multiple times before
+    and a reset the BPS if it reaches 0. The task can be called multiple times before
     the timer resets.
 
 Priority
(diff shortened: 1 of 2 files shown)


Docs/source/Software/Tasks.rst
```diff
@@ -187,6 +187,38 @@ Additional Considerations
     All possible CAN messages that will be sent to the rest of the car's system by the BPS
     are listed on the `CAN Bus IDs spreadsheet <https://docs.google.com/spreadsheets/d/11YWoMVZw8BFr8kyO4DIz0g-aIU_vVa0d-WioSRq85TI/edit#gid=0>`_.
 
+Pet WatchDog Task: Harshita Gorla & Clark Poon
```
ClarkPoon, line 190:
```suggestion
Pet WatchDog Task: Harshitha Gorla & Clark Poon
```

Docs/source/Software/Tasks.rst
```diff
@@ -187,6 +187,38 @@ Additional Considerations
...
+==============================================
+
+Purpose
+    The purpose of this task is to recognize if the BPS RTOS has stalled. If it is stuck somewhere
+    in the code, the car must shut down.
+
+Functionality
+    This task checks the ``WDog_BitMap`` variable to see if the 3 LSB are set by the temperature,
+    voltage, balancing, and current tasks. If these tasks ran and set those bits, that means that the BPS
+    is functional and the timer is reset. If those bits are not set, the timer will keep running
```
ClarkPoon, line 200: The fault handler is only called if the watchdog timer counts down to 0. The watchdog task may be called multiple times before the timer resets

Docs/source/Software/Tasks.rst
```diff
@@ -187,6 +187,38 @@ Additional Considerations
...
+    and a fault handler will be called.
+
+Priority
+    This task is priority 3. It is above the voltage, temperature, balancing, and current monitoring tasks
+    because if they run before the WatchDog timer is reset, the BPS will fault even if it is 
+    working correctly.
```
ClarkPoon, line 206: More specifically, we do not want this thread to be blocked by the other periodic threads, since the watchdog timer will reset the BPS if this thread is not called. TBH I am not sure how much it matters that this thread has a higher priority than the threads it is monitoring, since they will also cause the watchdog timer to reset the car if they are blocked for too long

Docs/source/Software/Tasks.rst
```diff
@@ -187,6 +187,39 @@ Additional Considerations
...
+    BatteryBalancing tasks.
+
+Timing Requirements
+    This task is set to run every 40 milliseconds.
+
+Yields
+    It doesn't yield.
+
+Additional Considerations
+    If we add more tasks(or split up tasks such as voltage and temperature) and want to have the
```
ClarkPoon, line 220:
```suggestion
    If we add more tasks (or split up tasks such as voltage and temperature) and want to have the 
```

Docs/source/Software/Tasks.rst
```diff
@@ -187,6 +187,39 @@ Additional Considerations
...
+Purpose
+    The purpose of this task is to recognize if the BPS RTOS has stalled. If it is stuck somewhere
+    in the code, the car must shut down.
+
+Functionality
+    This task checks the ``WDog_BitMap`` variable to see if the 3 LSB are set by the temperature,
+    voltage, balancing, and current tasks. If these tasks ran and set those bits, that means that the 
+    BPS is functional and the timer is reset. If those bits are not set, the timer will keep running
+    and a fault handler will be called if it reaches 0. The task can be called multiple times before
+    the timer resets.
```
ClarkPoon, line 202: One thing that is not clear from reading this is that the watchdog timer resets the BPS when it reaches zero and sets a specific bit to indicate that the watchdog timer reset the BPS. On startup, the BPS should check if that bit is set, and enters a fault state if it is. I can't find this check in the code right now, though, so we should add it in.

ClarkPoon (changes requested):

Docs/source/Software/Tasks.rst
ClarkPoon: I opened #312 for this

ClarkPoon (approved):

PR #308, opened 2021-01-30
Sending Charging Suggestions 

### LINKED ISSUE #306: Send charging suggestion over CAN
Whenever the highest temperature in the battery pack is between 45 Celsius and 60 Celsius, it is safe to discharge the battery but not to charge the battery. Whenever the highest temperature measured in the battery pack gets close to this range, the BPS should send a CAN message to suggest to the other systems that they should not charge the battery, or the BPS may shut down the car.

sugamarora99:
Fixes #306

### DIFF
--- a/Apps/Src/Task_VoltTempMonitor.c
+++ b/Apps/Src/Task_VoltTempMonitor.c
@@ -90,14 +90,6 @@ void Task_VoltTempMonitor(void *p_arg) {
                         OS_OPT_POST_1,
                         &err);
             assertOSError(err);
-            //suggest that the battery should not be charged
-            CanMsg.id = CHARGE_ENABLE;
-            CanPayload.idx = 0;
-            CanData.b = 0;
-            CanPayload.data = CanData;
-            CanMsg.payload = CanPayload;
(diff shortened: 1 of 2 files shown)


Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -90,6 +90,13 @@ void Task_VoltTempMonitor(void *p_arg) {
                         OS_OPT_POST_1,
                         &err);
             assertOSError(err);
+            //suggest that the battery should not be charged
+            CanMsg.id = CHARGE_ENABLE;
+            CanPayload.idx = 0;
+            CanData.b = 0;
+            CanPayload.data = CanData;
+            CanMsg.payload = CanPayload;
+            OSQPost(&CANBus_MsgQ, &CanMsg, sizeof(CanMsg), OS_OPT_POST_FIFO, &err);
```
ClarkPoon, line 99: This code will never run because posting the fault semaphore will call the Fault State task and kill all of the systems that are using this message (controls and array). We should instead be sending this signal when the highest temperature recorded in the battery box is between 45C and 60C (possibly with some tolerance on the lower end)

ClarkPoon (changes requested):

ClarkPoon (approved):

PR #317, opened 2021-02-13
Fault State issue

manthanand:
Fixed issue where fault state calls other tasks functions that could lead to yielding


Apps/Src/Amps.c
```diff
@@ -145,6 +145,7 @@ void Task_AmperesMonitor(void *p_arg) {
         // Check if amperes is NOT safe:
 		SafetyStatus amperesStatus = Amps_CheckStatus(Amps_IsCharging());
 		if(amperesStatus != SAFE) {
+			Fault_BitMap = 8;
```
ClarkPoon, line 148: Can we use an `enum` or a `#define` for these?

Apps/Src/Task_FaultState.c
```diff
@@ -31,26 +31,27 @@ void EnterFaultState() {
...
-        BSP_Light_On(OCURR);
-        EEPROM_LogError(FAULT_HIGH_CURRENT);
-    }
-    if (BSP_WDTimer_DidSystemReset()){
-        BSP_Light_On(WDOG);
-        EEPROM_LogError(FAULT_WATCHDOG);
+    switch (Fault_BitMap){
+        case 1:
+            BSP_Light_On(UVOLT);
+            EEPROM_LogError(FAULT_LOW_VOLT);
```
ClarkPoon, line 37: `EEPROM_LogError()` yields because it calls the I2C BSP. I am working on fixing the I2C BSP on a separate branch though, so I can fix this. My idea is to set it up so we can force the system to use the bare metal version of the BSP if we are in a fault state. What do you think @manthanand @chaseblock @sugamarora99

Apps/Src/Task_FaultState.c
```diff
@@ -31,26 +31,27 @@ void EnterFaultState() {
...
+            EEPROM_LogError(FAULT_HIGH_TEMP);
+            break;
+        case 8:
+            BSP_Light_On(OCURR);
+            EEPROM_LogError(FAULT_HIGH_CURRENT);
+            break;
+        default:
+            BSP_Light_On(WDOG); 
+            EEPROM_LogError(FAULT_WATCHDOG);
+            break;
```
ClarkPoon, line 54: It is possible to reach this block without faulting because of the watchdog. For example, we could have a hard fault or an open wire

ClarkPoon (changes requested):

Apps/Src/Task_FaultState.c
manthanand: But we do not have any log errors defined for those things so I'm not sure what to do for that.

Apps/Src/Task_FaultState.c
manthanand: I think it's a good idea.

Apps/Src/Task_FaultState.c
ClarkPoon: Change this so the default isn't the watchdog

ClarkPoon (changes requested):

Apps/Src/Task_FaultState.c
```diff
@@ -31,36 +31,53 @@ void EnterFaultState() {
...
-    }
-    if (BSP_WDTimer_DidSystemReset()){
-        BSP_Light_On(WDOG);
-        EEPROM_LogError(FAULT_WATCHDOG);
+    if (BSP_WDTimer_DidSystemReset()) Fault_BitMap = Fault_WDOG;
+    switch (Fault_BitMap){
+        case Fault_UVOLT:
+            BSP_Light_On(UVOLT);
+            EEPROM_LogError(FAULT_LOW_VOLT);
+            for(int i = 0; i < 31; i++) EEPROM_LogData(FAULT_LOW_VOLT, Voltage_GetModuleMillivoltage(i));
```
ClarkPoon, line 39: We have a `#define` for this in config.h

Apps/Inc/Tasks.h
```diff
@@ -119,5 +119,17 @@ extern OS_SEM SafetyCheck_Sem4;
...
+extern uint8_t Fault_Flag;
+typedef enum {
+    Fault_UVOLT   = 1,
+    Fault_OVOLT   = 2,
+    Fault_OTEMP   = 4,
+    Fault_OCURR   = 8,
+    Fault_OW      = 16,
+    Fault_HANDLER = 32,
+    Fault_OS      = 64,
+    Fault_WDOG    = 128,
```
ClarkPoon, line 132: I think hex is probably nicer for this, since we are treating this like a mask

Apps/Src/Task_FaultState.c
```diff
@@ -31,36 +31,53 @@ void EnterFaultState() {
...
+        case Fault_OVOLT:
+            BSP_Light_On(OVOLT);
+            EEPROM_LogError(FAULT_HIGH_VOLT);
+            for(int i = 0; i < 31; i++) EEPROM_LogData(FAULT_HIGH_VOLT, Voltage_GetModuleMillivoltage(i));
+            break;
+        case Fault_OTEMP:
+            BSP_Light_On(OTEMP);
+            EEPROM_LogError(FAULT_HIGH_TEMP);
+            uint8_t* tempArray = Temperature_GetModulesInDanger();
+            for(int i = 0; i < 63; i++) EEPROM_LogData(FAULT_HIGH_TEMP, tempArray[i]);
```
ClarkPoon, line 50: Use the `#define`

Apps/Src/Tasks.c
```diff
@@ -63,6 +63,8 @@ OS_MUTEX WDog_Mutex;
  */
 cell_asic Minions[NUM_MINIONS];
 uint32_t WDog_BitMap = 0;
+uint32_t Fault_BitMap = 0; //This is a variable that sets certain bits based on what caused the fault
```
ClarkPoon, line 66: This is fine, but since we stop monitoring after we detect the first fault, we will only ever set one bit

BSP/STM32F413/Src/BSP_I2C.c
```diff
@@ -1,44 +1,10 @@
-/* Copyright (c) 2020 UT Longhorn Racing Solar */
+//* Copyright (c) 2020 UT Longhorn Racing Solar */
```
ClarkPoon, line 1:
```suggestion
/* Copyright (c) 2020 UT Longhorn Racing Solar */
```
Oops, this was me

ClarkPoon (changes requested):

Apps/Src/Tasks.c
manthanand: So are you saying that it doesn't have to be a bitmap but just a normal variable? If needed I can change it, but it seems like either way would work.

ClarkPoon: Yeah, that's what I'm saying. I don't think we need to change it though

ClarkPoon (approved):

ClarkPoon (approved):

PR #319, opened 2021-02-13
I2C polling

ClarkPoon:
I reverted the I2C BSP to use polling instead of interrupts. This will prevent the Log Info task from blocking any higher priority tasks while it is in an ISR. This will also make it so that the Fault state task does not yield when it writes to the EEPROM

### DIFF
--- a/BSP/STM32F413/Src/BSP_I2C.c
+++ b/BSP/STM32F413/Src/BSP_I2C.c
@@ -1,43 +1,9 @@
-/* Copyright (c) 2020 UT Longhorn Racing Solar */
+//* Copyright (c) 2020 UT Longhorn Racing Solar */
 
 #include "BSP_I2C.h"
 #include "stm32f4xx.h"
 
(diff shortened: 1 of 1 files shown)


BSP/STM32F413/Src/BSP_I2C.c
```diff
@@ -78,310 +45,282 @@ void BSP_I2C_Init(void) {
...
+		if(timeout_count > TIMEOUT_THRESHOLD) {
+			return ERROR;
+		}
 	}
+
+	I2C_AcknowledgeConfig(I2C3, ENABLE);
+
+	// First write address that you want to read from
+
+	writePoll:
```
manthanand, line 74: tis a shame

manthanand (approved):

BSP/STM32F413/Src/BSP_I2C.c
ClarkPoon: Lol I blame @chaseblock

PR #323, opened 2021-02-13
PEC Error Code Fix

sugamarora99:
I couldn't come up with anything better than adding "_wrap" to the end of each of the 6 functions for the function names.

### DIFF
--- a/Drivers/Src/LTC6811.c
+++ b/Drivers/Src/LTC6811.c
@@ -150,6 +150,7 @@ void LTC6811_rdcv_safe(uint8_t reg, // Controls which cell voltage register is r
     if(count == MAX_PEC_ERRORS){
       //trip BPS
       OSSemPost(&Fault_Sem4, OS_OPT_POST_1, &err);
+      assertOSError(err);
     }
     count++;
   }
@@ -172,6 +173,7 @@ void LTC6811_rdaux_safe(uint8_t reg, //Determines which GPIO voltage register is
     if(count == MAX_PEC_ERRORS){
       //trip BPS
       OSSemPost(&Fault_Sem4, OS_OPT_POST_1, &err);
+      assertOSError(err);
     }
     count++;
   }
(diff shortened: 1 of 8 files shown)


ClarkPoon:
> I couldn't come up with anything better than adding "_wrap" to the end of each of the 6 functions for the function names.

Maybe `_safe` would be better?  What do you think?

Drivers/Src/LTC6811.c
```diff
@@ -138,6 +138,125 @@ void LTC6811_Init(cell_asic *battMod){
...
+  OS_ERR err;
+  uint8_t count = 0;
+  while(LTC681x_rdcv(reg,total_ic,ic) > -1){
+    if(count == MAX_PEC_ERRORS){
+      //trip BPS
+      OSSemPost(&Fault_Sem4, OS_OPT_POST_1, &err);
+    }
+    count++;
+  }
+}
```
ClarkPoon, line 156: I think this will fault unless we drop `MAAX_PEC_ERRORS` in a row. Shouldn't it be something like
`while(LTC6811xrdcv(reg,total_ic,ic) != 0){`
 or
`while(LTC6811xrdcv(reg,total_ic,ic) == -1){`

ClarkPoon (changes requested):

Drivers/Src/LTC6811.c
```diff
@@ -138,6 +138,125 @@ void LTC6811_Init(cell_asic *battMod){
...
+                     uint8_t total_ic, // the number of ICs in the system
+                     cell_asic ic[] // Array of the parsed cell codes
+                    )
+{
+  OS_ERR err;
+  uint8_t count = 0;
+  while(LTC681x_rdcv(reg,total_ic,ic) == -1){
+    if(count == MAX_PEC_ERRORS){
+      //trip BPS
+      OSSemPost(&Fault_Sem4, OS_OPT_POST_1, &err);
```
ClarkPoon, line 152: You should call `AssertOSError(err)` after calling an OS function

ClarkPoon (changes requested):

ClarkPoon (approved):

PR #330, opened 2021-02-20
Rtos can

chaseblock:


ClarkPoon:
One issue with this that we still need to fix is that if we try to send messages to the fifo too quickly, they will be dropped. @chaseblock is planning on fixing this tomorrow

ClarkPoon (approved):

ClarkPoon (approved):

PR #337, opened 2021-02-27
Added UART and SPI Priority Info to Documentation

sugamarora99:


ClarkPoon (changes requested):
Can you write the priority numbers for the interrupts? I can see that information being useful if we add another interrupt in the future. Also, you should put a note that our interrupts (UART and SPI) should be higher priority than the OS interrupts (Systick and Pendsv)

ClarkPoon (approved):

PR #349, opened 2021-04-03
Rtos fans

ClarkPoon:
I have verified that this works on the hardware

### DIFF
--- a/BSP/STM32F413/Src/BSP_Fans.c
+++ b/BSP/STM32F413/Src/BSP_Fans.c
@@ -102,13 +102,13 @@ void BSP_Fans_Init(void){
 
 void BSP_Fans_Init(void){
     //Enable TIM3 clock
-    RCC_APB2PeriphClockCmd(RCC_APB1Periph_TIM3, ENABLE);
+    RCC_APB2PeriphClockCmd(RCC_APB2Periph_TIM8, ENABLE);
 
     //Configure TIM pins by configuring corresponding GPIO pins
-    GPIO_PinAFConfig(GPIOC, GPIO_PinSource6, GPIO_AF_TIM3);
-    GPIO_PinAFConfig(GPIOC, GPIO_PinSource7, GPIO_AF_TIM3);
-    GPIO_PinAFConfig(GPIOB, GPIO_PinSource14, GPIO_AF_TIM3);
-    GPIO_PinAFConfig(GPIOB, GPIO_PinSource15, GPIO_AF_TIM3);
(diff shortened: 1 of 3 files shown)


BSP/STM32F413/Src/BSP_Fans.c
```diff
@@ -11,58 +11,70 @@ Uses Pins PC6,7 and PB14,15
...
+    GPIO_PinAFConfig(GPIOB, GPIO_PinSource14, GPIO_AF_TIM12);
+    GPIO_PinAFConfig(GPIOB, GPIO_PinSource15, GPIO_AF_TIM12);
+    //All pins will be initialized for alternate function
+    GPIO_StructInit(&GPIO_INIT_STRUCT);
     GPIO_INIT_STRUCT.GPIO_Mode = GPIO_Mode_AF;
-    GPIO_INIT_STRUCT.GPIO_OType = GPIO_OType_PP; 
-    GPIO_INIT_STRUCT.GPIO_PuPd = GPIO_PuPd_DOWN;
-    GPIO_INIT_STRUCT.GPIO_Speed = GPIO_Speed_2MHz;
+    GPIO_INIT_STRUCT.GPIO_PuPd = GPIO_PuPd_UP;
+    GPIO_INIT_STRUCT.GPIO_Speed = GPIO_Speed_50MHz;
```
manthanand, line 35: Is it necessary to have this high of a clock speed when the actual PWM speed is in the kHz?

BSP/STM32F413/Src/BSP_Fans.c
```diff
@@ -107,10 +119,10 @@ int BSP_Fans_GetSpeed(uint8_t fan){
         return TIM8-> CCR2 / DIVIDER;
         break;
     case 3:
-        return TIM8-> CCR3 / DIVIDER;
+        return TIM12-> CCR1 / DIVIDER;
```
manthanand, line 122: TIM_GetCapture4(TIMx) performs the same function but is provided by stm. Should probably replace this with that function instead.

BSP/STM32F413/Src/BSP_Fans.c
ClarkPoon: I am not sure. I used 50MHz because that's what some example code I found used. I can test this

BSP/STM32F413/Src/BSP_Fans.c
ClarkPoon: Will do

manthanand (approved):

PR #352, opened 2021-04-25
Working Interrupt SPI BSP

chaseblock:

PR #355, opened 2021-05-15
Working SPI and also some fixes for SPI3

ClarkPoon:
This includes all of the changes in #352 with some additional fixes and the test cases used to test the AS8510. Unfortunately, @chaseblock and I found an issue with the Amperes board hardware, so the AS8510 driver has not been verified

### DIFF
--- a/Config/Inc/config.h
+++ b/Config/Inc/config.h
@@ -20,7 +20,7 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 
 //--------------------------------------------------------------------------------
 // Basic Parameters of BPS layout
-#define NUM_MINIONS	1					 // Number of minion boards
+#define NUM_MINIONS	4					 // Number of minion boards
 
 //--------------------------------------------------------------------------------
 // Battery Fault Checks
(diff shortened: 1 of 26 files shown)


Config/Inc/config.h
```diff
@@ -20,7 +20,7 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 
 //--------------------------------------------------------------------------------
 // Basic Parameters of BPS layout
-#define NUM_MINIONS	4					 // Number of minion boards
+#define NUM_MINIONS	1					 // Number of minion boards
```
chaseblock, line 23: This is fine for testing, but shouldn't be committed.

Let's make sure that the config.h file stays proper for a production build.

chaseblock (changes requested):

Config/Inc/config.h
ClarkPoon: Good catch

JimothyGreene (approved):
All looks good to me. One question would just be why we need all the variants of test files for SPI.

ClarkPoon:
> All looks good to me. One question would just be why we need all the variants of test files for SPI.

They are for:
`Test_BSP_SP`I: testing `SPI1` when compiling for bare-metal
`Test_BSP_SPI_RTOS`: testing `SPI1` when compiling for Micrium
`Test_BSP_SPI3`: testing `SPI3` when compiling for Micrium

chaseblock (approved):

PR #357, opened 2021-05-15
Fault State Task safety-critical portions verified

ClarkPoon:
I verified that the fault state task turns off the contactor, turns the fans on at full speed, and turns on the correct LEDs. I have not checked if the EEPROM is logging the correct messages or if we are sending the correct CAN messages when it faults.

### DIFF
--- a/BSP/STM32F413/Src/BSP_Lights.c
+++ b/BSP/STM32F413/Src/BSP_Lights.c
@@ -3,6 +3,33 @@
 #include "BSP_Lights.h"
 #include "stm32f4xx.h"
 
+// Just some lookup for the values
+static const uint32_t LIGHT_VALS[LIGHTS_MAX] = {
+	GPIO_Pin_4,	// RUN, 
+	GPIO_Pin_7, // UVOLT, 
+	GPIO_Pin_6,	// OVOLT, 
+	GPIO_Pin_5,	// OTEMP, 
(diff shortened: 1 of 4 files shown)


BSP/STM32F413/Src/BSP_Lights.c
```diff
@@ -11,32 +38,21 @@
...
+	GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;	
+	GPIO_InitStruct.GPIO_PuPd =  GPIO_PuPd_NOPULL;
 	GPIO_InitStruct.GPIO_Speed = GPIO_Speed_50MHz;
 	GPIO_Init(GPIOC, &GPIO_InitStruct);
-	
-	// PA0 : Run
-	// PA5 : Extra
-	// PA6 : CAN
-	// PA7 : WDog
+
```
chaseblock, line 48: You know, since we have the arrays of pin numbers and ports anyways, I feel like we could write a more general initialization function.

Certainly not necessary though.

chaseblock (approved):
Looks fine to me. How did you test the fans? Just with a voltmeter, or with a few of them actually spinning?

ClarkPoon:
> Looks fine to me. How did you test the fans? Just with a voltmeter, or with a few of them actually spinning?

I did most of my testing with a voltmeter, but I did get an actual fan to spin at a different speed at each of the four fan ports on the fan board

PR #370, opened 2021-06-04
Test program for CAN task

ClarkPoon:
This adds a test program to test the CAN task. I also increased the length of the CAN message queue from 64 to 128 because I realized the VoltTemp task will send up to 31 + 1 + 62 = 94 CAN messages at a time.


JimothyGreene (approved):
Seems good to me

PR #376, opened 2021-06-26
Adds test for battery balancing task

JimothyGreene:
This test requires a lot from the actual people testing. If there are any suggestions on how to make it more automated or easier on y'all, just lmk.

### DIFF
--- a/Tests/Test_BatteryBalancing_Task.c
+++ b/Tests/Test_BatteryBalancing_Task.c
@@ -1,0 +1,63 @@
+/* Copyright (c) 2021 UT Longhorn Racing Solar */
+
+#include "BatteryBalancing.h"
+#include "Tasks.h"
+
+/******************************************************************************
+ * BatteryBalancing Task Test Plan
+ * 
+ * 1. Build the BPS code for the stm32f413 for this test file (see README.md)
+ * 2. Flash the BPS (see README.md)
+ * 3. Connect the BPS UART USB to your laptop and open PuTTY (or comparable)
+ * 4. Connect the BPS test board
+ * 5. Keep all modules at half voltage except one (of your choice) and keep that at full
+ * 6. Reset the BPS by pressing and releasing the reset button
+ * 7. Probe the balancing transistor to ensure it is turned on
+ * 8. Repeat 5-7 for different modules
+ ****************************************************************************/
+
+void Task1(void* p_arg) {
+	OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
+    OS_ERR err;
+
+    OSTaskCreate(
+        &BatteryBalance_TCB,				// TCB
+        "TASK_BATTERY_BALANCE",	// Task Name (String)
+        Task_BatteryBalance,				// Task function pointer
+        (void *)0,				// Task function args
+        6,			            // Priority
+        BatteryBalance_Stk,				// Stack
+        WATERMARK_STACK_LIMIT,	// Watermark limit for debugging
+        DEFAULT_STACK_SIZE,		// Stack size
+        0,						// Queue size (not needed)
+        10,						// Time quanta (time slice) 10 ticks
+        (void *)0,				// Extension pointer (not needed)
+        OS_OPT_TASK_STK_CHK | OS_OPT_TASK_SAVE_FP,	// Options
+        &err
+    );					// return err code}
+}
+
+void main(void) {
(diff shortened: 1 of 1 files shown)


Tests/Test_BatteryBalancing_Task.c
```diff
@@ -0,0 +1,63 @@
...
+/******************************************************************************
+ * BatteryBalancing Task Test Plan
+ * 
+ * 1. Build the BPS code for the stm32f413 for this test file (see README.md)
+ * 2. Flash the BPS (see README.md)
+ * 3. Connect the BPS UART USB to your laptop and open PuTTY (or comparable)
+ * 4. Connect the BPS test board
+ * 5. Keep all modules at half voltage except one (of your choice) and keep that at full
+ * 6. Reset the BPS by pressing and releasing the reset button
+ * 7. Probe the balancing transistor to ensure it is turned on
```
ClarkPoon, line 15: What is the best way to do this? Measure across the source to drain with an Ohm-meter? Measure the gate with a volt meter?

Tests/Test_BatteryBalancing_Task.c
```diff
@@ -0,0 +1,63 @@
...
+#include "Tasks.h"
+
+/******************************************************************************
+ * BatteryBalancing Task Test Plan
+ * 
+ * 1. Build the BPS code for the stm32f413 for this test file (see README.md)
+ * 2. Flash the BPS (see README.md)
+ * 3. Connect the BPS UART USB to your laptop and open PuTTY (or comparable)
+ * 4. Connect the BPS test board
+ * 5. Keep all modules at half voltage except one (of your choice) and keep that at full
```
ClarkPoon, line 13: We should also probably test if the BPS can balance multiple modules at the same time. This could be achieved by either setting one module to (at least) 50 mV less than the other modules

Tests/Test_BatteryBalancing_Task.c
```diff
@@ -0,0 +1,63 @@
...
+ * 6. Reset the BPS by pressing and releasing the reset button
+ * 7. Probe the balancing transistor to ensure it is turned on
+ * 8. Repeat 5-7 for different modules
+ ****************************************************************************/
+
```
ClarkPoon, line 18: We will also need to actually read voltages from the minion boards by calling `Voltage_UpdateMeasurements()`, since the battery balancing task uses `Voltage_GetModuleMillivoltage()` to access voltage readings. `Voltage_GetModuleMillivoltage()` does not read from the minion boards. It reads from an array in the STM32F413's memory that holds the most recently read voltage data.

ClarkPoon (changes requested):

Tests/Test_BatteryBalancing_Task.c
sugamarora99: In the event that this test doesn't work, I think we should consider refactoring the Battery Balancing library so that the Balancing_Balance() code could go into the thread and the helper functions could become non-static. This way testing each helper function from the Balancing library would become more convenient.

sugamarora99:
In the event that this test doesn't work, I think we should consider refactoring the Battery Balancing library so that the Balancing_Balance() code could go into the thread and the helper functions could become non-static. This way testing each helper function from the Balancing library would become more convenient.

ClarkPoon:
> In the event that this test doesn't work, I think we should consider refactoring the Battery Balancing library so that the Balancing_Balance() code could go into the thread and the helper functions could become non-static. This way testing each helper function from the Balancing library would become more convenient.

I think this would also make the code a bit more intuitive (since the setup would be ore similar to Voltage.c or Amps.c). In hindsight, the battery balancing code should also probably be in Apps instead of Drivers

Tests/Test_BatteryBalancing_Task.c
```diff
@@ -12,10 +15,31 @@
...
+// Task_UpdateVoltage
+OS_TCB UpdateVoltage_TCB;
+CPU_STK UpdateVoltage_Stk[DEFAULT_STACK_SIZE];
+
+void Task_UpdateVoltage(void* p_arg) {
+    (void) p_arg;
+    OS_ERR err;
+
+    while (1) {
+        Voltage_UpdateMeasurements();
```
ClarkPoon, line 37: Make sure to call `Voltage_Init()` before you call this

ClarkPoon (changes requested):

ClarkPoon (approved):

PR #379, opened 2021-06-26
Voltage Test Plan

ClarkPoon:
This creates the test plan for Voltage.c. It also makes some minor changes to other parts of the code, such as changing the types of a couple variables to types that make more sense and deleting an unused enum

### DIFF
--- a/Tests/Test_Voltage.c
+++ b/Tests/Test_Voltage.c
@@ -4,11 +4,115 @@
 #include "config.h"
 #include "Voltage.h"
 #include "BSP_UART.h"
-#include <unistd.h>
+#include "Tasks.h"
+#include "BSP_PLL.h"
+#include "stm32f4xx.h"
+#include "Amps.h"
+
+/******************************************************************************
+ * Voltage App Test Plan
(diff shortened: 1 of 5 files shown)


Tests/Test_Voltage.c
```diff
@@ -84,3 +188,36 @@ int main() {
 
     while(1);
 }
+
+// Task for this test
+void Task1(void *p_arg){
+	OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
+	
+    test();
+}
```
JimothyGreene, line 197: Do we need to put a blank while loop here? I'm not sure how this will act without an idle task to fall back on

JimothyGreene (commented):
Seems mostly fine to me. Just one small question

Tests/Test_Voltage.c
ClarkPoon: Added this just now to the end of `test()`

JimothyGreene (approved):

PR #391, opened 2021-07-31
remove most floating point math and disable floating point unit

ClarkPoon:


Apps/Src/Charge.c
```diff
@@ -34,17 +34,14 @@ void Charge_Calculate(int32_t milliamps){
...
-	float accumulatedAmpHrs = timeElapsed * fAmps;
-	
+	uint32_t ticksElapsed = 0xFFFF - counter;	// I hate this, but fixing BSP_Timer_GetTicksElapsed() is beyond the scope of this PR. Opened issue #389 for this
+
+	uint32_t clockFrequency = BSP_Timer_GetRunFreq();
+
 	/* Update Charge, units of 0.01% */
-	charge += (uint32_t) (CHARGE_RESOLUTION_SCALE * (100 * (accumulatedAmpHrs / MAX_CHARGE_AMP_HRS)));
+	// TODO: I am preserving the existing math to make this PR easier to follow. Fixing it is a problem for issue #390
+	charge += (uint32_t) (CHARGE_RESOLUTION_SCALE * 100 * ticksElapsed * milliamps
```
chaseblock, line 43: Damn this is ugly.

Apps/Src/Task_Amperes.c
```diff
@@ -45,7 +46,7 @@ void Task_AmperesMonitor(void *p_arg) {
 
 		//Send measurement to CAN queue
 		int current = Amps_GetReading();
-		CanData.f = (float)current/1000; //send data in Amps
+		memcpy(&(CanData.w), &current, sizeof(CanData.w));
```
chaseblock, line 49: Is there any reason in particular to be doing a memcpy right here, instead of an assignment?

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -56,7 +57,7 @@ void Task_VoltTempMonitor(void *p_arg) {
         for (int i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module voltage data
             CanPayload.idx = i;
             int voltage = Voltage_GetModuleMillivoltage(i);
-            CanData.f = (float)voltage/1000; //send data in volts
+            memcpy(&CanData.w, &voltage, sizeof(CanData.w));
```
chaseblock, line 60: same question as before.

Apps/Src/Temperature.c
```diff
@@ -155,15 +163,24 @@ ErrorStatus Temperature_ChannelConfig(uint8_t tempChannel) {
  * @param mV from ADC
  * @return temperature in Celsius (Fixed Point with .001 resolution) 
  */
-int milliVoltToCelsius(float milliVolt){
+int milliVoltToCelsius(uint32_t milliVolt){
+	// Adapted from: 
+	/*
 	// Typecasting to get rid of warnings
```
chaseblock, line 169: Could probably get rid of this comment.

Apps/Src/Temperature.c
```diff
@@ -155,15 +163,24 @@ ErrorStatus Temperature_ChannelConfig(uint8_t tempChannel) {
...
-	float retVal = frac + 30;			
-	
+	float retVal = frac + 30;	
+	*/
+
+	uint32_t sumInRt = (184471 + 17 * (2230800 - milliVolt * 1000)) / 1000;
+	uint32_t rt = uintSqrt(sumInRt);
+	int32_t numerator = 13582000 - rt * 1000000;
+	int32_t denominator = -8660;
+	int32_t retVal = numerator / denominator + 30;
```
chaseblock, line 182: Might be good to get some idea of how much precision we're losing from this, and whether there's any other order of operations that would buy us more precision.

Apps/Src/Task_Amperes.c
ClarkPoon: `CanData.w` is unsigned, but `current` is signed, so I don’t think we can typecast safely if we have a negative `current`

Apps/Src/Task_VoltTempMonitor.c
ClarkPoon: Same reason as before, but we could probably make `voltage` unsigned

Apps/Src/Charge.c
ClarkPoon: Do you think it would be better if I broke this up into multiple lines?

Apps/Src/Task_Amperes.c
ClarkPoon: I looked into this some more, and I think we can safely do this with a cast, so I'll change it

Apps/Src/Temperature.c
ClarkPoon: I replaced this with a lookup table because I think we have the flash space, and I don't want to worry about losing precision

chaseblock (dismissed):
That's a big LUT

manthanand (approved):

PR #395, opened 2021-08-29
CAN bug fixes

ClarkPoon:
I fixed the following bugs (verified on hardware using Test_BSP_CAN, Test_RTOS_CAN, and Test_CAN_Task):
- Fixed initialization of CAN Tx interrupts in BSP
- Allowed BSP_CAN and CANBus to be initialized in loopback mode
- Updated list of valid CAN messages and associated switch statement
- Fixed CANBus error handling that was causing resource leak
- Changed some defines in config.h to the correct values

### DIFF
--- a/Apps/Src/CAN_Queue.c
+++ b/Apps/Src/CAN_Queue.c
@@ -1,0 +1,63 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+/** CAN_Queue.c
+ * Queue that holds all CAN messages that Task_CANBusConsumer needs to send.
+ */
+
+#include "CAN_Queue.h"
+#include "os.h"
+#include "CANbus.h"
+#include "Tasks.h"
+
+// fifo
+#define FIFO_TYPE CANMSG_t
+#define FIFO_SIZE 256
+#define FIFO_NAME CAN_fifo
+#include "fifo.h"
+
+static CAN_fifo_t canFifo;
+static OS_SEM canFifoSem;
+static OS_MUTEX canFifoMutex;
+
+void CAN_Queue_Init(void) {
+    OS_ERR err;
+    CPU_TS ticks;
+    OSMutexCreate(&canFifoMutex, "CAN queue mutex", &err);
+    assertOSError(err);
+    OSSemCreate(&canFifoSem,
+                "CAN queue semaphore",
+                0,
+                &err);
+    assertOSError(err);
+    OSMutexPend(&canFifoMutex, 0, OS_OPT_POST_NONE, &ticks, &err);
+    assertOSError(err);
+    CAN_fifo_renew(&canFifo);
+    OSMutexPost(&canFifoMutex, OS_OPT_POST_NONE, &err);
+    assertOSError(err);
+}
+
+void CAN_Queue_Post(CANMSG_t message) {
+    OS_ERR err;
+    CPU_TS ticks;
+    OSMutexPend(&canFifoMutex, 0, OS_OPT_POST_NONE, &ticks, &err);
+    assertOSError(err);
+    CAN_fifo_put(&canFifo, message);
+    OSMutexPost(&canFifoMutex, OS_OPT_POST_NONE, &err);
+    assertOSError(err);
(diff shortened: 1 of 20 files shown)


Apps/Src/CAN_Queue.c
```diff
@@ -0,0 +1,63 @@
...
+    assertOSError(err);
+    CAN_fifo_renew(&canFifo);
+    OSMutexPost(&canFifoMutex, OS_OPT_POST_NONE, &err);
+    assertOSError(err);
+}
+
+void CAN_Queue_Post(CANMSG_t message) {
+    OS_ERR err;
+    CPU_TS ticks;
+    OSMutexPend(&canFifoMutex, 0, OS_OPT_POST_NONE, &ticks, &err);
```
manthanand, line 41: Place comments here for what to do if CAN_fifo_put returns error.

Apps/Src/CAN_Queue.c
```diff
@@ -0,0 +1,63 @@
...
+#include "Tasks.h"
+
+// fifo
+#define FIFO_TYPE CANMSG_t
+#define FIFO_SIZE 256
+#define FIFO_NAME CAN_fifo
+#include "fifo.h"
+
+static CAN_fifo_t canFifo;
+static OS_SEM canFifoSem;
```
manthanand, line 18: Should we have these follow the same naming system as others? Name_Sem4, Name_Mutex

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -3,6 +3,7 @@
 #include "BSP_CAN.h"
 #include "stm32f4xx.h"
 #include "os.h"
+#include "BSP_Lights.h"
```
manthanand, line 6: Should this be removed if it was only used for testing purposes?

Config/Inc/config.h
```diff
@@ -54,14 +54,14 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
...
-#define TOTAL_VOLT_WIRES					NUM_PINS_PER_LTC*NUM_MINIONS - 1
+#define MAX_VOLT_SENSORS_PER_MINION_BOARD	8	// User defined. The LTC6811 can actually measure 12 modules.
+#define NUM_PINS_PER_LTC 					(MAX_VOLT_SENSORS_PER_MINION_BOARD + 1) // one extra for ground
+#define TOTAL_VOLT_WIRES					(NUM_PINS_PER_LTC * NUM_MINIONS - 1)
 
 //--------------------------------------------------------------------------------
 // Temperature Sensor Configurations
 // Define how many temperature sensors are connected to each board
-#define MAX_TEMP_SENSORS_PER_MINION_BOARD	12
+#define MAX_TEMP_SENSORS_PER_MINION_BOARD	(MAX_VOLT_SENSORS_PER_MINION_BOARD * NUM_TEMP_SENSORS_PER_MOD)
```
manthanand, line 64: I think this was supposed to stay at 12.

Apps/Src/CAN_Queue.c
ClarkPoon: Changed this to return an `ErrorStatus` as discussed in-person

Apps/Src/CAN_Queue.c
ClarkPoon: Yeah, I agree it's a good idea to keep the naming consistent. These aren't new shared resources though because they are internal to this module. The module itself is a shared resource that is replacing a different shared resource (the CAN OS queue)

BSP/STM32F413/Src/BSP_CAN.c
ClarkPoon: yes, good catch

Config/Inc/config.h
ClarkPoon: The correct value is 16. I changed this to 12 a few months ago during testing and forgot to change it back. This number is because there are 2 temperature sensors on each module, and each board connects to 8 modules (except for the last one, which connects to 7)

manthanand (approved):

PR #407, opened 2021-10-16
Volt temp monitor task test

manthanand:


Tests/Test_VoltTempMonitorTask.c
```diff
@@ -0,0 +1,162 @@
...
+            10,						// Time quanta (time slice) 10 ticks
+            (void *)0,				// Extension pointer (not needed)
+            OS_OPT_TASK_STK_CHK | OS_OPT_TASK_SAVE_FP,	// Options
+            &err);					// return err code
+        
+    // Initialize CAN queue
+    CAN_Queue_Init();
+    assertOSError(err);
+
+    // get contactor to close without temperature or voltage readings
```
ClarkPoon, line 124: Why?

Tests/Test_VoltTempMonitorTask.c
```diff
@@ -0,0 +1,162 @@
...
+    OSSemCreate(&SafetyCheck_Sem4,
+                "Safety Check Semaphore",
+                0,
+                &err);
+    assertOSError(err);
+
+    OSMutexCreate(&WDog_Mutex,
+                "Watchdog Mutex",
+                &err);
+    assertOSError(err);
```
ClarkPoon, line 54: You will also need a task to periodically set the watchdog bitmap for the amperes task and the battery balancing task

ClarkPoon (changes requested):

ClarkPoon:
I verified this on the BPS hardware with power supplies and breadboards instead of temperature sensors and battery modules. I had 6 "modules" and 12 "temperature sensors".

I disabled the open wire checks because they slow the task down by several seconds. This causes the watchdog to time out before the task can complete a single loop. It would be nice to have open wire checks working within our time constraints, but I don't think they are as safety critical as checking voltage and temperature because the voltage and temperature checks usually catch any open wires with a second or so (based on my experiments).

The task takes around 250 ms to run (profiled by looking at SPI traffic on logic analyzer). We might want to look into making this more efficient in the future, but I don't think it is critical. This limits our sample rate to 4 Hz max, but with the 100 ms delay at the end of the task, we will realistically hit something closer to 3 Hz.

ClarkPoon:
I forgot to mention this in my earlier comment, but the CAN consumer task is also not getting enough CPU time to send all of the CAN messages it wants to. CAN isn't safety critical, so I think we can push this off to later

Apps/Src/Task_FaultState.c
```diff
@@ -74,15 +74,19 @@ void EnterFaultState() {
...
-    CANPayload_t Message;
-    Message.data = data;
-    // Push Trip message to CAN Q
-    CANbus_BlockAndSend(TRIP, Message);
-    // Push Contactor State message to CAN Q
-    data.b = 0;
-    CANbus_BlockAndSend(CONTACTOR_STATE, Message);
+
+    // avoid infinite recursive faults, since our CAN Driver relys on the OS to work
+    if (Fault_BitMap != Fault_OS) {
```
chaseblock, line 79: Is it possible for the fault bitmap to be in two fault states at once? I.e. should this be `(Fault_BitMap & Fault_OS) == 1`?

Apps/Src/Task_PetWDog.c
```diff
@@ -30,7 +30,7 @@ void Task_PetWDog(void *p_arg) {
         OSMutexPost(&WDog_Mutex, OS_OPT_POST_NONE, &err);
         assertOSError(err);
 
-        OSTimeDly(20, OS_OPT_TIME_DLY, &err);
+        OSTimeDly(10, OS_OPT_TIME_DLY, &err);
```
chaseblock, line 33: Any particular reason for this change?

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -66,7 +67,10 @@ void Task_VoltTempMonitor(void *p_arg) {
 
         // BLOCKING =====================
         // Check if open wire is NOT safe:
-        SafetyStatus wireStatus = Voltage_OpenWire();
+	
+	// TODO: get OpenWire to not take forever and actually run this
```
chaseblock, line 71: Open an issue ticket

Apps/Src/Voltage.c
```diff
@@ -194,6 +194,9 @@ SafetyStatus Voltage_OpenWire(void){
 
 	for(int32_t i = 0; i < NUM_MINIONS; i++) {
 		if(Minions[i].system_open_wire != 0){
+			if ((i == NUM_MINIONS -1) && ((Minions[i].system_open_wire & 0xEF) != 0)) { //The last Voltage board is only connected to 7 modules
```
chaseblock, line 197: Could we make this dependent on the configuration?

Apps/Src/Task_FaultState.c
```diff
@@ -74,15 +74,19 @@ void EnterFaultState() {
...
+        CANData_t data;
+        data.b = 1;
+        CANPayload_t Message;
+        Message.data = data;
+        // Push Trip message to CAN Q
+        CANbus_BlockAndSend(TRIP, Message);
+        // Push Contactor State message to CAN Q
+        data.b = 0;
+        CANbus_BlockAndSend(CONTACTOR_STATE, Message);
+    }
```
manthanand, line 89: So if it is an OS Fault, does this mean that we can't send a TRIP message over CAN? Is there something else we should do instead?

Apps/Src/Task_PetWDog.c
```diff
@@ -30,7 +30,7 @@ void Task_PetWDog(void *p_arg) {
...
         assertOSError(err);
```
manthanand, line 34: Was this decreased because we are not running the open wire checks anymore? Would there be anything wrong with leaving it at 200ms?

Apps/Src/Voltage.c
manthanand: I think we could do that by adding more defines to the config file (like batteries missing in last board or something like that).

Apps/Src/Task_FaultState.c
ClarkPoon: We can still log the error to the EEPROM.

On other thing I noticed that I didn't get around to adding in this pull request is that there is no field in the Fault_BitMap for if we fault by dropping too many SPI packets in a row. I can open a ticket for this

Apps/Src/Task_PetWDog.c
ClarkPoon: I don't think it affects that much. 100 ms makes it use more CPU time, but makes us less likely to timeout the watchdog when we shouldn't

Apps/Src/Task_VoltTempMonitor.c
ClarkPoon: Opened #416

Apps/Src/Voltage.c
ClarkPoon: Referenced this in #416

Apps/Src/Task_FaultState.c
ClarkPoon: Opened #417 and #418 for the issues mentioned above

ClarkPoon (approved):
Approving based on my own testing/review and @chaseblock 's verbal approval

PR #409, opened 2021-11-20
Bsp timert bugs

kevingonz19:
Timer.c and Charge.c are both altered. Timer.c was changed to use an upcounter and a prescaler to allow .1 seconds to count 2000 ticks. This is used in Get_Ticks_Elapsed to return the proper number of ticks elapsed between function calls. I then added a function called Get_Micros_Elapsed which converts the ticks from the previous function into microseconds and returns that. The use of microseconds is to avoid precision loss within our math in future functions added and modified in this pull request. Additionally, Clark changed Timer_GetRunFreq to return the frequency used by the timer, since the original code I worked on returned a different frequency which ruined the math in returning micro seconds.

Charge.c was altered to use a larger charge resolution scale, to have a more accurate max mAh measurement, and to use unsigned int milliamps instead of floats within the charge calculation to prevent precision loss. Now, within the Charge_Calculate function, charge is decremented in units of .000001%. Finally, there is some type casting to 64 bit unsigned integers within the previous function to prevent overflow issues we had while debugging.

Both functions were tested using UART and the OS sleep function. In our final stage of debugging we used a 100 millisecond interval which at a timer frequency of 40,000,00Hz and a prescaler of 2000 gives us 2000 ticks. Inputting this into the getMicros function returns 100,000 microseconds, which is correct for time between function calls. The rest of the math in Charge_Calculate using 100k microseconds and a reading of 5000 mA returned the proper value to decrement charge by, where 100% charge = 100,000,000.

### DIFF
--- a/Apps/Src/Charge.c
+++ b/Apps/Src/Charge.c
@@ -34,10 +34,10 @@ void Charge_Calculate(int32_t milliamps){
 	
 	int64_t micro_sec = (int64_t)BSP_Timer_GetMicrosElapsed();
 	int64_t millis = (int64_t)milliamps;
-	
-	charge -= (int32_t) (micro_sec * millis
-						* ((100 * CHARGE_RESOLUTION_SCALE) / 1000000)
-						/ 3600 / MAX_CHARGE_MILLI_AMP_HRS);
+	                                                                   // Psuedo code that represents what this math does. In the coded math, order of operations is very important to avoid overflow 
+	charge -= (int32_t) (micro_sec * millis                            // microseconds / 1,000,000 / 3600 = hours_elapsed
+						* ((100 * CHARGE_RESOLUTION_SCALE) / 1000000)  // (milliamp * hours_elapsed) / max_amp_hour = percent_of_charge_elapsed
+						/ 3600 / MAX_CHARGE_MILLI_AMP_HRS);            // percent_charge_elapsed * 100 * charge resolution = a much larger number which represents the charge depleted
 }
 
 /** Charge_Calibrate
--- a/Apps/Src/Charge.c
+++ b/Apps/Src/Charge.c
@@ -26,17 +26,17 @@ void Charge_Init(void){
 
 /** Charge_Calculate
  * Calculates new Charge depending on the current reading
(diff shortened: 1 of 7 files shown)


Apps/Src/Charge.c
```diff
@@ -27,21 +26,18 @@ void Charge_Init(void){
...
-
-	uint32_t clockFrequency = BSP_Timer_GetRunFreq();
-
-	/* Update Charge, units of 0.01% */
-	// TODO: I am preserving the existing math to make this PR easier to follow. Fixing it is a problem for issue #390
-	charge += (uint32_t) (CHARGE_RESOLUTION_SCALE * 100 * ticksElapsed * milliamps 
-							/ clockFrequency / 1000 / 60 / 60 / 60 / MAX_CHARGE_AMP_HRS);
+	charge -= (int32_t) (micro_sec * millis
+						* ((100 * CHARGE_RESOLUTION_SCALE) / 1000000)
+						/ 3600 / MAX_CHARGE_MILLI_AMP_HRS);
```
ClarkPoon, line 40: Can you put a comment to explain what is going on in this line?

BSP/STM32F413/Src/BSP_Timer.c
```diff
@@ -13,9 +16,9 @@ void BSP_Timer_Init(void) {
 	RCC_APB1PeriphClockCmd(RCC_APB1Periph_TIM2, ENABLE);
 	TIM_TimeBaseInitTypeDef Init_TIM2;
 	
-	Init_TIM2.TIM_Prescaler = 1;
-	Init_TIM2.TIM_CounterMode = TIM_CounterMode_Down;
-	Init_TIM2.TIM_Period = 0xFFFF-1;
+	Init_TIM2.TIM_Prescaler = PRESCALER;
+	Init_TIM2.TIM_CounterMode = TIM_CounterMode_Up;
+	Init_TIM2.TIM_Period = 0xFFFF;
```
ClarkPoon, line 21: Can you add a comment that explains how this affects the clock timing? It would be good to know what the clock frequency is and the maximum time we can reliably measure (how long it takes for the clock to count down from `0xFFFF` to `0`).

ClarkPoon (approved):

PR #421, opened 2021-12-21
Config Parameters for Open Wire Test

manthanand:
Added parameters to config file so changes in battery module configuration do not require changes in Voltage.c file.


Config/Inc/config.h
```diff
@@ -33,6 +33,8 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 #define NUM_BATTERY_MODULES				31		// Number of battery modules
 #define NUM_TEMPERATURE_SENSORS			62		// Number of temperature sensors
 #define NUM_TEMP_SENSORS_PER_MOD		2		// Number of temperature sensors per battery module
+#define NUM_BATTERY_MODULES_MISSING     1       // Number of battery modules less than maximum value per minion board
```
ClarkPoon, line 36: I think it would be better to define this in terms of `MAX_VOLT_SENSORS_PER_MINION_BOARD` (it might make sense to rename this to something like `NUM_BATTERY_MODULES_PER_MINION`) and `NUM_BATTERY_MODULES`, so we don't have to change the config file in as many places if we change the pack configuration

Config/Inc/config.h
```diff
@@ -33,6 +33,8 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
...
+#define OPEN_WIRE_MASK                  0xEF    // Mask used to ignore disconnected battery modules in Open Wire Test
```
ClarkPoon, line 37: I don't understand the calculation that led to this number. I would think that `0x7F` would make more intuitive sense than `0xEF` (`b01111111` instead of `b11101111`), but I haven't looked at how the open wire code actually works.

ClarkPoon:
Why are there 28 commits in this pull request? I think it would make more sense to just rebase your one commit to be in a new branch based on `rtos` instead of `VoltTempMonitorTaskTest`

PR #422, opened 2021-12-22
Config Parameters for Open Wire Test

manthanand:
I implemented two ways to create the mask. One way is to just have the mask as a config parameter and the other is to calculate the mask based on other config parameters during the initialization function.

### DIFF
--- a/Apps/Src/Voltage.c
+++ b/Apps/Src/Voltage.c
@@ -63,10 +63,16 @@ void Voltage_Init(cell_asic *boards){
   	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
   	assertOSError(err);
 
-	for (int i = 0; i < NUM_BATTERY_MODULES_PER_MINION; i++){
-		open_wire_mask <<= 1;
-		open_wire_mask |= 1;
+	// set the lower bits of open_wire_mask to 1, such that there are as many 1s are modules connected to the last minion
+	open_wire_mask = 1;
+	if  (NUM_BATTERY_MODULES % NUM_BATTERY_MODULES_PER_MINION) {
+		// handle case where the last minion has a different number of modules than the other minions
+		open_wire_mask <<= (NUM_BATTERY_MODULES % NUM_BATTERY_MODULES_PER_MINION) + 1;
+	} else {
+		// handle case where the last minion has the same number of modules as the other minions
+		open_wire_mask <<= NUM_BATTERY_MODULES_PER_MINION + 1;
 	}
+	open_wire_mask -= 1;
 	open_wire_mask >>= NUM_BATTERY_MODULES_MISSING; 
 }
 
--- a/Apps/Src/Voltage.c
+++ b/Apps/Src/Voltage.c
@@ -86,7 +86,7 @@ void Voltage_UpdateMeasurements(void){
 	OSMutexPend(&Voltage_Mutex, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
   	assertOSError(err);
 	for(int i = 0; i < NUM_BATTERY_MODULES; i++){
-		VoltageVal[i] = Minions[i / MAX_VOLT_SENSORS_PER_MINION_BOARD].cells.c_codes[i % MAX_VOLT_SENSORS_PER_MINION_BOARD];
+		VoltageVal[i] = Minions[i / NUM_BATTERY_MODULES_PER_MINION].cells.c_codes[i % NUM_BATTERY_MODULES_PER_MINION];
 	}
 	//release mutex
   	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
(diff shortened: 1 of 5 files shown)


manthanand:
Also changed MAX_VOLT_SENSORS_PER_MINION_BOARD to NUM_BATTERY_MODULES_PER_MINION since it can be used outside of voltage.

Apps/Src/Voltage.c
```diff
@@ -61,7 +62,12 @@ void Voltage_Init(cell_asic *boards){
 	// release mutex
   	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
   	assertOSError(err);
-	
+
+	for (int i = 0; i < NUM_BATTERY_MODULES_PER_MINION; i++){
+		open_wire_mask <<= 1;
+		open_wire_mask = 1;
```
ClarkPoon, line 68:
```suggestion
		open_wire_mask |= 1;
```

Config/Inc/config.h
```diff
@@ -33,6 +33,10 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 #define NUM_BATTERY_MODULES				31		// Number of battery modules
 #define NUM_TEMPERATURE_SENSORS			62		// Number of temperature sensors
 #define NUM_TEMP_SENSORS_PER_MOD		2		// Number of temperature sensors per battery module
+#define NUM_BATTERY_MODULES_PER_MINION	8	// User defined. The LTC6811 can actually measure 12 modules.
+// Number of battery modules less than maximum value per minion board
+#define NUM_BATTERY_MODULES_MISSING     NUM_BATTERY_MODULES - ((NUM_MINIONS-1) * NUM_BATTERY_MODULES_PER_MINION)
+#define OPEN_WIRE_MASK                  0x7F    // Mask used to ignore disconnected battery modules in Open Wire Test
```
ClarkPoon, line 39: Is this used anywhere?

ClarkPoon (commented):
Are you planning on testing these changes on the BPS before you merge?

Config/Inc/config.h
manthanand: I am thinking that we only need line 38 or 39. If we use 38, then we need that code in the init function above. Or we can use line 39 and get rid of that code that creates the mask. I personally prefer having line 39 and getting rid of line 38 because it would be easier to understand later on.

Apps/Src/Voltage.c
```diff
@@ -18,6 +18,7 @@ static cell_asic *Minions;
 static OS_MUTEX Voltage_Mutex;
 static uint16_t VoltageVal[NUM_BATTERY_MODULES]; //Voltage values gathered
 static uint32_t openWires[TOTAL_VOLT_WIRES];
+static long open_wire_mask = 0;
```
ClarkPoon, line 21: Why is this a `long` and not `uintXX_t`? 

Also, why `open_wire_mask` instead of `openWireMask`? (I realize our variable naming conventions are all over the place in this file, so I don't care too much about this)

Apps/Src/Voltage.c
```diff
@@ -61,7 +62,12 @@ void Voltage_Init(cell_asic *boards){
 	// release mutex
   	OSMutexPost(&MinionsASIC_Mutex, OS_OPT_POST_NONE, &err);
   	assertOSError(err);
-	
+
+	for (int i = 0; i < NUM_BATTERY_MODULES_PER_MINION; i++){
+		open_wire_mask <<= 1;
+		open_wire_mask |= 1;
+	}
```
ClarkPoon, line 69:
```suggestion
	// set the lower NUM_BATTERY_MODULES_PER_MINION bits of open_wire_mask to 1
	open_wire_mask = 1;
	open_wire_mask <<= NUM_BATTERY_MODULES_PER_MINION + 1;
	open_wire_mask -= 1;
```
I think this will accomplish the same thing more efficiently (except in the edge cases where `NUM_BATTERY_MODULES_PER_MINION > 31`, which is not supported by our hardware anyway). Not sure if this is worth it if it reduces the code readability too much, since it is only called once.

Config/Inc/config.h
ClarkPoon: With the addition of `open_wire_mask` to Voltage.c, shouldn't we remove `OPEN_WIRE_MASK`?

Apps/Src/Voltage.c
ClarkPoon: After looking at this some more, the open wire mask is only being used for the last minion board, so we should be generating this based on `NUM_BATTERY_MODULES % NUM_BATTERY_MODULES_PER_MINION` instead of `NUM_BATTERY_MODULES_PER_MINION`. For example:

```suggestion
	// set the lower bits of open_wire_mask to 1, such that there are as many 1s are modules connected to the last minion
	open_wire_mask = 1;
	open_wire_mask <<= (NUM_BATTERY_MODULES % NUM_BATTERY_MODULES_PER_MINION) + 1;
	open_wire_mask -= 1;

ClarkPoon:
I think this should be retargeted to the master branch

Apps/Src/Voltage.c
manthanand: I made it a long because the data type it is being "&" with was a long, `Minions[i].system_open_wire`.

Config/Inc/config.h
manthanand: Removed it

ClarkPoon (commented):
Looks good to me. We should probably test this though

PR #426, opened 2022-01-03
modified amperes checks to take the battery pack temperature into account

### LINKED ISSUE #425: The original intent here was to have different charging cutoffs based on the highest temperature measured in the battery pack. If any cells are above `MAX_CHARGE_TEMPERATURE_LIMIT` (defined in `Config/Inc/config.h`), the BPS should not allow charging. I am realizing now that we are using this improperly, and this bug slipped by the original reviews and testing, so good catch. I will make a fix for this
The original intent here was to have different charging cutoffs based on the highest temperature measured in the battery pack. If any cells are above `MAX_CHARGE_TEMPERATURE_LIMIT` (defined in `Config/Inc/config.h`), the BPS should not allow charging. I am realizing now that we are using this improperly, and this bug slipped by the original reviews and testing, so good catch. I will make a fix for this

_Originally posted by @ClarkPoon in https://github.com/lhr-solar/BPS/pull/271#discussion_r777265413_

ClarkPoon:
fixes #425

### DIFF
--- a/Apps/Src/Amps.c
+++ b/Apps/Src/Amps.c
@@ -88,19 +88,19 @@ ErrorStatus Amps_UpdateMeasurements(void) {
  * @return SAFE or DANGER
  */
 SafetyStatus Amps_CheckStatus(int32_t maxTemperature) {
-	bool chargingOnly = maxTemperature > MAX_CHARGE_TEMPERATURE_LIMIT;
 	OS_ERR err;
 	CPU_TS ticks;
 	SafetyStatus status;
+
+	// determine if we should allow charging or not
+	bool dischargingOnly = maxTemperature > MAX_CHARGE_TEMPERATURE_LIMIT;
+	int32_t chargingCurrentLimit = dischargingOnly ? 0 : MAX_CHARGING_CURRENT;
+
 	OSMutexPend(&AmperesData_Mutex, 0, OS_OPT_PEND_BLOCKING, &ticks, &err);
 	assertOSError(err);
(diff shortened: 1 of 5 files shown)


Apps/Src/Amps.c
```diff
@@ -84,11 +84,11 @@ ErrorStatus Amps_UpdateMeasurements(void) {
...
 /** Amps_CheckStatus
  * Checks if pack does not have a short circuit
- * @param chargingOnly a flag that indicates whether we should allow
- * 		  only charging or both charging and discharging.
+ * @param maxTemperature the maximum temperature measured in the most recent batch of temperature readings
  * @return SAFE or DANGER
  */
-SafetyStatus Amps_CheckStatus(bool chargingOnly) {
+SafetyStatus Amps_CheckStatus(int32_t maxTemperature) {
+	bool chargingOnly = maxTemperature > MAX_CHARGE_TEMPERATURE_LIMIT;
```
manthanand, line 91: Why is it just MAX_CHARGE_TEMPERATURE_LIMIT? Wouldn't we have to check the sign of the current to see whether it is the charge or discharge temperature limit?

manthanand, line 91: Would this mean we want to only charge when the max temperature is larger than the MAX_CHARGE_TEMPERATURE_LIMIT? Shouldn't it be the other way around, only discharge if it is greater than this limit (since the charge limit is 45C and the discharge limit is 60C)?

ClarkPoon: Yeah, you're right. I'll fix this

manthanand (approved):

PR #432, opened 2022-01-10
Automated testing skeleton

ClarkPoon:
I think it would be neat if we could get a few tests as part of our automated checks. I made some minor fixes to the simulator to get it to compile, and added a basic test for the CAN driver as a proof of concept. I'm interested to see what y'all think about this.

Please let me know any questions, comments, suggestions, etc.

A few things I would like feedback on are:
- Do you think we can get meaningful test coverage out of this without a ton of extra work? I will try to make a test for one of voltage/temperature/amperes later to try this
- Does it look like the process for adding tests is straightforward? Can we reasonably expect all of our members to understand how this works?
- I did not put a whole lot of thought into directory/file structure or naming, so any feedback would be appreciated
- Anything else you want to comment on


ClarkPoon:
I also removed the timer from the BPS simulator because it uses a ton of CPU and is not a good simulation anyway. I am not sure if these changes break the interactive simulator's CAN implementation, but I don't think the interactive simulator is very useful anyway.

.github/workflows/verify-build.yml
```diff
@@ -17,9 +17,13 @@ jobs:
         sudo apt -y install gcc-arm-none-eabi
     - name: Initialize submodules
       run: git submodule update --init --recursive
-#    - name: make simulator
-#      run: make simulator
-    - name: clean
-      run: make clean
     - name: make stm32f413
```
manthanand, line 20: Is there where we could change it so the default is "make stm32f413" but all we have to type is "make"?

chaseblock (approved):
Looks good to me. May want to get some others to look at this before merging though.

.github/workflows/verify-build.yml
```diff
@@ -17,9 +17,13 @@ jobs:
...
       run: make stm32f413
+    - name: clean
+      run: make clean
+    - name: make simulator
+      run: make simulator
+    - name: build CANbus
+      run: rm ./bps-simulator.out && make simulator TEST=CI_CANbus && mkdir BSP/Simulator/DataGeneration/Data
+    - name: run CANbus
+      run: sudo ./Continuous_Integration/CANbus/Test_CANbus.sh
```
manthanand, line 29: Should we add documentation to this PR on how to add this here (maybe in the Readme)?

PR #433, opened 2022-01-10
Small updates to README and Docs

ClarkPoon:
updated rules on code reviews. re-oredered approvers based on how many reviews I think they will do in the near future. Updated shunt resistor and fuse docs


chaseblock (approved):
Looks fine. May want to consider expanding the list of approved reviewers soon though.

manthanand (approved):

PR #435, opened 2022-01-15
Check contactor task

ClarkPoon:
The BPS faults now if it detects that the contactor is open more than 250 ms after it closes the contactor, so the strobe and fault indicator will turn on if the emergency stop button is pressed


BSP/STM32F413/Src/BSP_Contactor.c
```diff
@@ -22,7 +22,7 @@ void BSP_Contactor_Init(void) {
 	GPIO_InitStruct.GPIO_Mode = GPIO_Mode_OUT;
 	GPIO_InitStruct.GPIO_Speed = GPIO_Speed_50MHz;
 	GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;
-	GPIO_InitStruct.GPIO_PuPd = GPIO_PuPd_DOWN;
+	GPIO_InitStruct.GPIO_PuPd = GPIO_PuPd_UP;
```
manthanand, line 25: Why does the logic of the Contactor change now that we are checking if it is open or not?

manthanand:
In this PR, should there be a test function to see if the Contactor task correctly detects the change? Also, considering timing and priority, should this be combined with another task (such as amperes) or kept as its own task?

BSP/STM32F413/Src/BSP_Contactor.c
ClarkPoon: This was wrong before. It was correct on the master branch, but wrong on the rtos branch, and I picked the wrong one when I resolved the merge conflict.

This pin is connected to one of the contactor sense pins. The other contactor sense pin is connected to ground. When the contactors are closed, the contactors will short the contactor sense pins together, which ties this pin to ground externally. When the contactors are open, this pin will be floating externally, so we need the internal pull-up resistor to make this pin high when the contactors are open

ClarkPoon:
> In this PR, should there be a test function to see if the Contactor task correctly detects the change? Also, considering timing and priority, should this be combined with another task (such as amperes) or kept as its own task? Also, should the WDOG_Bitmap variable be updated to include this task if we are planning on having it as it's own task?

I tested this with Test_CriticalState_1.c
 
I don't see the point of combining this with another task. Are you suggesting that this should be higher priority? I think that would make sense (eg. make it higher than the tasks that take measurements)

I don't think the watchdog needs to worry about this task. I think the only reason this task would get stuck is if higher priority tasks are hogging the CPU.

I'll update the priority for this task to be higher than volt/temp

ClarkPoon:
I also updated the instructions for the critical state test plan to include the test I did today

manthanand (approved):

PR #438, opened 2022-02-14
Integration bugs

ClarkPoon:


Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -111,13 +111,23 @@ void Task_VoltTempMonitor(void *p_arg) {
...
+                }
             }
         }
+        // for(uint8_t k = 0; k < NUM_BATTERY_MODULES; k++){
+        //     uint32_t temp = Temperature_GetModuleTemperature(k);
+        //     if(temp > MAX_CHARGE_TEMPERATURE_LIMIT && temp < MAX_DISCHARGE_TEMPERATURE_LIMIT){
+        //         //suggest that the battery should not be charged
+        //         charge_enable = false;
+        //     }
+        // }
```
manthanand, line 130: Don't need this anymore, do we?

Apps/Src/Temperature.c
```diff
@@ -22,6 +23,14 @@ static uint8_t ChargingState;
...
 static cell_asic *Minions;
 
+void Temperature_delay_m(uint16_t milli)
+{
+  uint32_t delay = BSP_PLL_GetSystemClock() / 1000;
+	for(uint32_t i = 0; i < milli; i++)
+	{
+		for(uint32_t j = 0; j < delay; j++);
+	}
+}
```
manthanand, line 33: make static?

Apps/Src/Temperature.c
```diff
@@ -168,6 +174,8 @@ ErrorStatus Temperature_UpdateSingleChannel(uint8_t channel){
 	if (ERROR == Temperature_ChannelConfig(channel)) {
 		return ERROR;
 	}
+
+	// wait for channel to stabilize
```
manthanand, line 178: Was the temperature delay function supposed to be here? If we don't need it anymore we can remove these comments.

Apps/Src/Task_VoltTempMonitor.c
ClarkPoon: I can remove this

Apps/Src/Temperature.c
ClarkPoon: I'll check if it's used anywhere and remove it if it's not

Apps/Inc/Temperature.h
```diff
@@ -22,6 +22,8 @@
 #define AUX_I2C_BLANK 0
 #define AUX_I2C_NO_TRANSMIT 0x7
 
+#define TEMP_ERR_OUT_BOUNDS 42424242
```
HannahSLee14, line 25: Where did 42424242 come from?

Apps/Src/Task_FaultState.c
```diff
@@ -77,7 +81,9 @@ void EnterFaultState() {
     }
 
     // avoid infinite recursive faults, since our CAN Driver relies on the OS to work
-    if ((Fault_BitMap & Fault_OS) == 0) {
+    // also don't call CAN if the watchdog tripped, since CAN won't be initialized
+    /*
+    if ((Fault_BitMap & (Fault_OS | Fault_WDOG)) == 0) {
```
HannahSLee14, line 86: If this is all commented, is there a purpose to keeping it around? Is it just in case we need it in the future?

Apps/Src/Task_VoltTempMonitor.c
```diff
@@ -111,13 +111,23 @@ void Task_VoltTempMonitor(void *p_arg) {
         }
         bool charge_enable = true;
         //Check if car should be allowed to charge or not
-        for(uint8_t k = 0; k < NUM_BATTERY_MODULES; k++){
-            uint32_t temp = Temperature_GetModuleTemperature(k);
-            if(temp > MAX_CHARGE_TEMPERATURE_LIMIT && temp < MAX_DISCHARGE_TEMPERATURE_LIMIT){
-                //suggest that the battery should not be charged
-                charge_enable = false;
+        for (uint8_t board = 0; board < NUM_MINIONS; board++) {
```
HannahSLee14, line 114: Is the purpose of breaking the single for loop into a nested for loop just to make it clearer what the actual system (ie 4 minions, 8 modules for each except for 1 with 7) is like, or is there another purpose?

Apps/Src/Temperature.c
```diff
@@ -225,12 +240,26 @@ ErrorStatus Temperature_UpdateAllMeasurements(){
...
+		// for example, if there are 14 sensors, use sensor numbers (1-indexed) 1-7 and 9-15
+		// int numChannels = (i + 1 < NUM_MINIONS) ? MAX_TEMP_CHANNELS : CHANNELS_IN_LAST_MINION;
+		// for (int j = 0; j < numChannels; j++) {
 		for (int j = 0; j < MAX_TEMP_SENSORS_PER_MINION_BOARD; j++) {
 			if (i * MAX_TEMP_SENSORS_PER_MINION_BOARD + j >= NUM_TEMPERATURE_SENSORS) break;
-			if (ModuleTemperatures[i][j] > temperatureLimit) {
+			if ((temperatures[i][j] > temperatureLimit) || (temperatures[i][j] == TEMP_ERR_OUT_BOUNDS)) {
 				return DANGER;
 			}
+			/*
```
HannahSLee14, line 258: Same question as before - since all of this is commented, are we keeping this around just in case we need it in the future?

Apps/Src/Temperature.c
HannahSLee14, line 33: I think I'm missing this somewhere but where do we use this? I noticed the Temperature_delay_m hasn't been added to the header file and none of the functions in Temperature.c seem to be using it

Apps/Inc/Temperature.h
ClarkPoon: Its an impossibly high temperature (42424.242 degrees C) that we return as an error code if we get a voltage reading from the temperature sensor that is out of bounds of what the datasheet says the LMT87 will return (eg. 5 Volts). 

The motivation for this is:
1. The temperature sensors will never actually measure ~42,000 C
2. This number is high enough to cause an overvoltage fault
3. When printed out during debugging, it looks made up to a human

Apps/Src/Task_VoltTempMonitor.c
ClarkPoon: Yeah, I think it's easier to read and `Temperature_GetSingleTempSensor()` takes `board` and `sensor` as inputs, so I think it's a little bit nicer if we already have those values from the nested loop variables

Apps/Src/Temperature.c
ClarkPoon: I'll take this out. We aren't using it anymore

Apps/Src/Temperature.c
ClarkPoon: I don't think we're using this either. We added it in while debugging, and I never got around to removing it. I'll get rid of it

Apps/Src/Task_FaultState.c
HannahSLee14: We can keep it in if we might use it in the future!

Apps/Src/Temperature.c
```diff
@@ -228,7 +232,7 @@ SafetyStatus Temperature_CheckStatus(uint8_t isCharging){
 	for (int i = 0; i < NUM_MINIONS; i++) {
 		for (int j = 0; j < MAX_TEMP_SENSORS_PER_MINION_BOARD; j++) {
 			if (i * MAX_TEMP_SENSORS_PER_MINION_BOARD + j >= NUM_TEMPERATURE_SENSORS) break;
-			if (ModuleTemperatures[i][j] > temperatureLimit) {
+			if ((temperatures[i][j] > temperatureLimit) || (temperatures[i][j] == TEMP_ERR_OUT_BOUNDS)) {
```
manthanand, line 235: Do we need to have it explicitly check for TEMP_ERR_OUT_BOUNDS or is it just for readability?

BSP/Inc/BSP_SPI.h
```diff
@@ -25,7 +25,7 @@ typedef enum {
  * @param   spi_os pointer to struct that holds the SPI specific pend()/post() functions 
  * @return  None
  */
-void BSP_SPI_Init(spi_port_t port, bsp_os_t *spi_os);
+void BSP_SPI_Init(spi_port_t port, bsp_os_t *spi_os, bool baremetal);
```
manthanand, line 28: Has the bare-metal code been tested? Do we want to keep it if we can do without it for now or are we saving it for another case?

Apps/Src/Temperature.c
ClarkPoon: It's for readability and in case someone ever makes a code change that causes `TEMP_ERR_OUT_BOUNDS` to be a safe voltage

Drivers/Src/LTC2315.c
```diff
@@ -58,22 +58,29 @@ void LTC2315_Sleep() {
...
+    LTC2315_wakeup_sleep();
 
     uint8_t count = 0;
     do {
-    LTC2315_wakeup_sleep();
+    OSSchedLock(&err);
+    assertOSError(err);
 
     BSP_SPI_SetStateCS(spi_ltc2315, 0);
     BSP_SPI_Read(spi_ltc2315, rxdata, 2);
```
manthanand, line 71: Should we add a note here that if we send more than 8 bytes, the OS will fault.

BSP/STM32F413/Makefile
```diff
@@ -9,7 +9,7 @@ DEBUG = 1
 # building variables
 ######################################
 # optimization
-OPT = -Og
+OPT = -O3
```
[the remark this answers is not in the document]
ClarkPoon: I'll make a change so it runs with -Og if DEBUG is 1 and -O3 otherwise

Drivers/Src/LTC2315.c
ClarkPoon: I can add the comment

BSP/STM32F413/Makefile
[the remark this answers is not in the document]
ClarkPoon: On second thought, I think a comment would be better because I don't think changing the `DEBUG` variable is significantly easier than changing the `OPT` variable

PR #440, opened 2022-02-27
added UART logging to BSP_CAN

ClarkPoon:
This is the modified version of the BPS code I want to run for the fully day test on Wednesday. I don't actually plan on merging this (unless we want to enable something like this in debug mode), but I'm opening a pull request to get some code reviews


manthanand:
I think it would be nice if we could have something like this in debugging mode (at least until CLI is finished). Could we just surround this in an ifdef in case we wanted to leave it in master?

ClarkPoon:
This was throttling the system when we tried to run it on test day because UART is slower than CAN, and we send a lot of CAN messages, so I think we should not do this

PR #441, opened 2022-04-02
Added state of charge to amperes thread and rewrote EEPROM code

ClarkPoon:
I haven't tested this with the real main yet, so opening it as a draft


ClarkPoon:
I got it working

Apps/Src/CLI.c
```diff
@@ -1,809 +1,809 @@
...
-					printf("Invalid ID\n\r");
-					break;
-			}
-			break;
-		default:
-			printf("Invalid CAN command\n\r");
-			break;
-	}
-}
+// /* Copyright (c) 2020 UT Longhorn Racing Solar */
```
manthanand, line 1: Is there a  reason all of this is commented out?

Apps/Src/Task_LogInfo.c
```diff
@@ -8,11 +8,10 @@ void Task_LogInfo(void *p_arg) {
     (void)p_arg;
 
     OS_ERR err;
-    EEPROM_Init();
 
     while (1){
         //sends percentage of charge left in the battery pack to EEPROM
-        EEPROM_LogData(DATA_SOC, Charge_GetPercent());
+        EEPROM_SetCharge(Charge_GetPercent());
```
manthanand, line 14: Is it necessary to have this in another task? Would it make more sense to send the charge data in the amps task but have a counter that counts down the number of times the Amps task runs before the EEprom is written to(to prevent too many writes)? Also should EEPROM_Init be run here?

manthanand: I feel like we would get issues with being interrupted in the middle of a write. According to datasheet fastest speed is around 400kHz.

Apps/Src/CLI.c
ClarkPoon: It doesn't compile with the new EEPROM library. I can fix the compile errors if you want

Apps/Src/Task_LogInfo.c
ClarkPoon: The advantage of putting this in its own task is that if the EEPROM function takes too long or gets stuck, it won't affect the amperes sampling rate. I am guessing that this function call will take at least 15ms because I think the EEPROM needs to wait 5ms in between byte writes, but we can profile it if we want.

Yeah, we might have issues when we are interrupted. I can add in error handling and see if any of the I2C messages are failing

manthanand (approved):

PR #453, opened 2022-05-05
Median filtering

ClarkPoon:
Fix for issue where regen braking will cause spikes in current that create noise on voltage and temperature measurement lines with median filtering.

Also, branched off of the EEPROM fix branch, so this includes the fixes from there

### DIFF
--- a/Apps/Src/Voltage.c
+++ b/Apps/Src/Voltage.c
@@ -13,6 +13,13 @@
 #include "Tasks.h"
 #include "Amps.h"
 
+// for median filter
+// if you change this, you also need to change the calls to median()
+#define VOLT_MEDIAN_FILTER_DEPTH 3
+
+static uint16_t rawVoltages[NUM_BATTERY_MODULES][VOLT_MEDIAN_FILTER_DEPTH];
+static int32_t voltFilterIdx = VOLT_MEDIAN_FILTER_DEPTH - 1;
+
 static cell_asic *Minions;
 
 static OS_MUTEX Voltage_Mutex;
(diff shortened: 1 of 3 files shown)


Apps/Src/Voltage.c
```diff
@@ -82,10 +113,21 @@ void Voltage_UpdateMeasurements(void){
...
+		rawVoltages[i][voltFilterIdx] = Minions[i / MAX_VOLT_SENSORS_PER_MINION_BOARD].cells.c_codes[i % MAX_VOLT_SENSORS_PER_MINION_BOARD];
+	}
+	voltFilterIdx = (voltFilterIdx + 1) % VOLT_MEDIAN_FILTER_DEPTH;
+
+	// update public voltage values
 	OSMutexPend(&Voltage_Mutex, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
   	assertOSError(err);
 	for(int i = 0; i < NUM_BATTERY_MODULES; i++){
-		VoltageVal[i] = Minions[i / MAX_VOLT_SENSORS_PER_MINION_BOARD].cells.c_codes[i % MAX_VOLT_SENSORS_PER_MINION_BOARD];
+		// you need to change this if you change VOLT_MEDIAN_FILTER_DEPTH
```
manthanand, line 127: I don't think it makes sense to have filter depth change since the actual filter function can't handle different values. Maybe remove these comments and just have a comment at the #define? If it could handle multiple values then we should consider a filter.h file.

ClarkPoon: I left this comment, so it will be easier to see if that this will break if we change the `#define`. There is also a comment at the `#define`. I agree that we should create a separate file for filters if we start to mess with them a lot.

manthanand (approved):

PR #460, opened 2022-05-28
Created generic median filter and used it in voltage.c

ClarkPoon:
Also increased the filter depth for voltage measurements
Also made some minor coding style improvements to voltage.c

This isn't tested yet.

MedianFilter.h is inspired by fifo.h


ClarkPoon:
Marking this as a draft until we can test it on the hardware

manthanand (approved):

PR #469, opened 2022-06-14
Added Charge disable for high voltage readings

manthanand:


Config/Inc/config.h
```diff
@@ -39,15 +39,16 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 #define MIN_VOLTAGE_CHARGING_LIMIT      2550        // Under voltage limit for charging (milliVolts) 
 #define MIN_VOLTAGE_LIMIT				2700		// Under voltage limit (milliVolts)	(actual min: 2.5V)
 #define MAX_VOLTAGE_LIMIT				4000		// Over voltage limit (milliVolts)		(actual max: 4.2V)
+#define CHARGE_DISABLE_VOLTAGE          3900        // Voltage to stop charging at
```
ClarkPoon, line 42: It might be better to define this relative to `MAX_VOLTAGE_LIMIT`, like `(MAX_VOLTAGE_LIMIT - 100)`, so if we change the max voltage limit, this will change too. 

Also, we can probably increase our voltage cutoff to something closer to 4.2 V (the manufacturer max) since we are cutting off charging earlier. Maybe we can increase to 4.1?

manthanand: I don't think we have to keep it the same distance every time so I will change the cutoffs but leave them independent of each other.

Config/Inc/config.h
```diff
@@ -38,16 +38,17 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 // MAX and MIN limits of lithium ion cells
 #define MIN_VOLTAGE_CHARGING_LIMIT      2550        // Under voltage limit for charging (milliVolts) 
 #define MIN_VOLTAGE_LIMIT				2700		// Under voltage limit (milliVolts)	(actual min: 2.5V)
-#define MAX_VOLTAGE_LIMIT				4000		// Over voltage limit (milliVolts)		(actual max: 4.2V)
+#define MAX_VOLTAGE_LIMIT				4100		// Over voltage limit (milliVolts)		(actual max: 4.2V)
+#define CHARGE_DISABLE_VOLTAGE          4000        // Voltage to stop charging at
```
ClarkPoon, line 42: Could also include something like
```C
#if MAX_VOLTAGE_LIMIT - 100 < CHARGE_DISABLE_VOLTAGE
#error "Charging maximum voltage is too close to voltage trip limit!"
#endif
```
Or do something with the `#pragma` you found

ClarkPoon (approved):

PR #473, opened 2022-08-27
Rtos library

Siddhant779:
Made the basic structure for the RTOS library files


Tasks/Inc/Task_RTOSBPS.h
```diff
@@ -0,0 +1,54 @@
...
+
+//#include
+
+
+/**
+ * @brief   
+ * @param   
+ * @return  
+ */
+void RTOS_SemPend(void);
```
manthanand, line 16: Could you rename to RTOS_BPS_blah. I want something that no other RTOS function would have in their name so I would like to include BPS to prevent compiler errors for future porting.

Siddhant779: 👍🏼

yuvrajkhullar (dismissed):
looks good

manthanand (approved):

PR #477, opened 2022-09-14
Wrapper for QueuePut

connorl309:
Currently following Micrium spec - wrapper for putting a message into an RTOS Queue

### DIFF
--- a/Tasks/Inc/BPS_Fifo.h
+++ b/Tasks/Inc/BPS_Fifo.h
@@ -1,0 +1,174 @@
+/**
+ *  Copyright Longhorn Racing Solar 2022-2023 
+ *  Implements a FIFO structure on Micrium's Heap
+ * 
+*/
+#ifndef __BPS_FIFO_H
+#define __BPS_FIFO_H
+
+// Includes
+#include "os.h"
+#include "lib_mem.h"
+#include <stdint.h>
+#include <stdbool.h>
+
+// Flags for the Fifo
+typedef enum {
+    FIFO_INITIALIZED,
+    FIFO_EMPTY,
+    FIFO_FULL,
+    FIFO_AVAILABLE,
+    FIFO_ERR
+} fifo_flags;
+
+// Fifo struct
+typedef struct {
+    size_t fifoSize;
+    size_t elementSize;
+    intptr_t* putptr;
+    intptr_t* getptr;
+    intptr_t* fifo; // range of fifo base address -> fifo base address[fifoSize - 1]
+    unsigned int flags;
+} BPS_Fifo_t;
+
+// An invalid fifo - returned on errors
+const BPS_Fifo_t INVALID_FIFO = {0, 0, NULL, NULL, NULL, FIFO_ERR};
+
+/**
+ * @brief Returns size of an element in the FIFO
+ * 
+ * @param fifo 
+ * @return size_t 
+ */
+inline size_t fifoTypeSize(BPS_Fifo_t* fifo) {
+    return fifo->elementSize;
+}
+
+/**
+ * @brief Determines if fifo is empty or non empty
+ * 
+ * @param fifo - pointer to a fifo struct (should be created with Fifo_Init)
+ * @return true if empty
+ * @return false if not empty
+ */
+inline bool fifoEmpty(BPS_Fifo_t* fifo) {
+    return fifo->putptr == fifo->getptr;
+}
+
+/**
+ * @brief Determines if fifo is full or not full
+ * 
+ * @param fifo - pointer to a fifo struct (should be created with Fifo_Init)
+ * @return true if full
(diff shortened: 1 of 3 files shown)


Tasks/Src/RTOS_BPS.c
```diff
@@ -25,6 +25,20 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+ * Note that the incoming BPS_OS_QUEUE queue !!MUST!! be created already.
+ * @brief   Puts a message into the RTOS Queue (i.e., POST something into the Queue). https://docs.silabs.com/micrium/latest/micrium-kernel-api/03-kernel-message-queue-api
+ * @param   *queue - pointer to the Queue to write a message to
+ * @param   *msg - void pointer to the message you are going to send
+ * @param   size - payload size of *msg (this changes in FreeRTOS)
+ * @return  none
+ */
+void RTOS_BPS_QueuePut(BPS_OS_QUEUE* queue, void* msg, BPS_OS_MSG_SIZE size) {
+    BPS_OS_ERR err;
+    OSQPost(queue, msg, size, OS_OPT_POST_FIFO, &err);
```
tiandahuang, line 39: is support for stack/LIFO planned?

connorl309: OS_OPT_POST_FIFO is stuff built into Micrium. If you mean using Queues overall, I think it might be used when we transition to FreeRTOS but I think Manthan would have a better idea about the broader scope of that. The wrapper exists for now as a more generic call for Queues and once we decide fully on when and how to use them it'll get some more updates to it.

Tasks/Src/RTOS_BPS.c
```diff
@@ -25,6 +25,20 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+
+/**
+ * Note that the incoming BPS_OS_QUEUE queue !!MUST!! be created already.
+ * @brief   Puts a message into the RTOS Queue (i.e., POST something into the Queue). https://docs.silabs.com/micrium/latest/micrium-kernel-api/03-kernel-message-queue-api
+ * @param   *queue - pointer to the Queue to write a message to
+ * @param   *msg - void pointer to the message you are going to send
+ * @param   size - payload size of *msg (this changes in FreeRTOS)
+ * @return  none
+ */
+void RTOS_BPS_QueuePut(BPS_OS_QUEUE* queue, void* msg, BPS_OS_MSG_SIZE size) {
```
manthanand, line 37: Specify this to FIFO. That is pretty much all we use for now.

Tasks/Src/RTOS_BPS.c
manthanand: We don't use this function anywhere in the BPS code. I would recommend looking at fifo.h and CAN_Queue.c. Don't change it yet though, because I'm not too sure why we aren't using this function.

tiandahuang: I thought we don't use micrium queues anywhere and instead use fifo.h + semaphore

manthanand: Yea, I remember there being a reason why, but I don't remember what it is. I would assume the queues are thread safe. fifo.h + semaphore should probably be used in this function.

connorl309:
Tried writing a new Fifo.h file. Uses Micrium heap. Trying a wacky way to create it with custom types, not explicitly specified in an array, but by doing casts on the returned pointer.

Tasks/Inc/BPS_Fifo.h
```diff
@@ -0,0 +1,174 @@
...
+// An invalid fifo - returned on errors
+const BPS_Fifo_t INVALID_FIFO = {0, 0, NULL, NULL, NULL, FIFO_ERR};
+
+/**
+ * @brief Returns size of an element in the FIFO
+ * 
+ * @param fifo 
+ * @return size_t 
+ */
+inline size_t fifoTypeSize(BPS_Fifo_t* fifo) {
```
manthanand, line 43: Probably don't need this PR anymore, but I would use static inline to prevent compiler optimization.

Tasks/Inc/BPS_Fifo.h
```diff
@@ -0,0 +1,174 @@
...
+        fifo->putptr -= fifo->elementSize;
+        // handle wraparound
+        if (fifo->putptr < fifo->fifo) 
+            fifo->putptr = (fifo->fifo + fifo->elementSize*fifo->fifoSize);
+        returnPtr = fifo->putptr;
+    }
+    return returnPtr;
+}
+
+#endif
```
manthanand, line 174: Just a note, try to add newlines to the end of every file.

PR #478, opened 2022-09-16
Implementing MutexCreate

connorl309:


Tasks/Src/RTOS_BPS.c
```diff
@@ -10,6 +10,20 @@ void RTOS_BPS_SemPend(void) {
...
 
 }
+
+/**
+ * @brief Creates and returns a mutex.
+ * @param name - char* of the name of the mutex
+ * @return an BPS_OS_MUTEX object
+ */
+BPS_OS_MUTEX RTOS_BPS_MutexCreate(char* name) {
+    BPS_OS_MUTEX mut;
```
tiandahuang, line 20: This creates the mutex as a local variable (on stack), not sure if you can return mutexes by value

connorl309: You're probably right, will change

PR #479, opened 2022-09-17
RTOS Mutex Pend Wrapper

ray4477:
Mutex Pend Wrapper for Micrium.
Takes in a pointer to Mutex, time in ticks to wait or 0 to wait indefinitely, and an option to block. 
Returns timestamp and error in pointers ticks and err respectively.

### DIFF
--- a/Tasks/Inc/RTOS_BPS.h
+++ b/Tasks/Inc/RTOS_BPS.h
@@ -47,7 +47,7 @@ void RTOS_BPS_MutexCreate(BPS_OS_MUTEX *mut, char* name);
  * @brief   Function waits for Mutex
  * @param   *mutex - pointer to mutex
  * @param   timeout - timeout period, if 0 will wait forever until resource available
- * @param   options - determines if mutex available or not
+ * @param   options - determines what the mutex will do, ie: block or not block
  */
 void RTOS_BPS_MutexPend(BPS_OS_MUTEX* mutex, BPS_OS_TICK timeout, BPS_OS_OPT opt);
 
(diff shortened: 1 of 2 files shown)


connorl309:
looks good

manthanand (approved):

manthanand:
Try fixing the merge conflicts on this PR.

Tasks/Src/RTOS_BPS.c
```diff
@@ -11,6 +11,20 @@ void RTOS_BPS_SemPost(void) {
 
 }
 
+/**
+ * @brief   Waits for Mutex, assigns timestamp and any error to err and ticks
+ * @param   *mutex - pointer to mutex
+ * @param   timeout - timeout period, if 0 will wait forever until resource available
+ * @param   options - determines if mutex available or not
```
manthanand, line 18: Is this comment correct?

ray4477: yes i think so

manthanand: no it aint, look at the options parameter in the micrium function.

Tasks/Inc/RTOS_BPS.h
```diff
@@ -44,11 +44,12 @@ void RTOS_BPS_SemPost(void);
 void RTOS_BPS_MutexCreate(BPS_OS_MUTEX *mut, char* name);
 
 /**
- * @brief   
- * @param   
- * @return  
+ * @brief   Function waits for Mutex
+ * @param   *mutex - pointer to mutex
+ * @param   timeout - timeout period, if 0 will wait forever until resource available
+ * @param   options - determines if mutex available or not
```
manthanand, line 50: You gotta change the comment in both files.

ClarkPoon:
Are you planning on changing the BPS code to call your function?

connorl309:
Yeah, the wrappers will get finalized and then I believe once we port over
to freertos or when manthan decides we will rewrite the bps code to use the
wrappers

PR #482, opened 2022-09-21
add RTOS_BPS_TaskCreate wrapper function

Champers5000:
- [x] Code Build checks pass 
- [x] No merge conflicts 


manthanand:
Rename this PR to something that more accurately describes what it is meant to do.

Tasks/Src/RTOS_BPS.c
```diff
@@ -38,9 +40,19 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+    CPU_STK_SIZE   stk_size,
+    OS_ERR        *p_err){
+        OSTaskCreate(p_tcb, p_name, p_task, p_arg, prio, p_stk_base, WATERMARK_STACK_LIMIT, stk_size, 0, 10,(void *)0, OS_OPT_TASK_STK_CHK | OS_OPT_TASK_SAVE_FP, p_err);
+        assertOSError(*p_err);
+        }
 
-}
 void RTOS_BPS_TimeDelay(void){
 
 }
```
manthanand, line 58: Add a newline to the end of all files. You can look at the coding styles guide on the BPS documentation for more information.

Tasks/Src/RTOS_BPS.c
```diff
@@ -38,9 +40,19 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
     OSMutexPost(mutex, options, &err);
     assertOSError(err);
 }
-void RTOS_BPS_TaskCreate(void){
+void RTOS_BPS_TaskCreate(
```
manthanand, line 43: Need comment headers.

manthanand:
When you push a commit, make sure the message you write accurately describes the commit to some extent.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -149,8 +149,11 @@ ErrorStatus BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
     
     gTxMessage.StdId = id;
     gTxMessage.DLC = length;
-	for(int i = 0; i < length; i++){
-        gTxMessage.Data[i] = data[i];
+
+    if(length<=8){
+        for(int i=0; i<length; i++){
+            gTxMessage.Data[i] = data[i];
+        }
```
manthanand, line 156: The general practice for PR's in GitHub is that one PR changes one issue in the code. I am guessing you did this to fix the gdb errors in the code which is an issue separate from the RTOS Create Task. It is fine for now, but for later on I just wanted you to know.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -149,8 +149,11 @@ ErrorStatus BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
     
     gTxMessage.StdId = id;
     gTxMessage.DLC = length;
-	for(int i = 0; i < length; i++){
-        gTxMessage.Data[i] = data[i];
+
+    if(length<=8){
```
manthanand, line 153: Is there a reason you chose 8? If so could you write it as a comment? There should be a reason there is no else statement following so if someone is looking at the code, they should know why we wouldn't send a CAN message if the size was less than 8. (Hint: if the reason is there is a max limit on CAN message sizes, leave it as a macro at the start of the file)

Tasks/Src/RTOS_BPS.c
```diff
@@ -4,7 +4,9 @@ RTOS_BPS library includes all the wrapper functions for a functioning RTOS - nee
 
 #include "RTOS_BPS.h"
 
-void RTOS_BPS_SemPend(void) {
+#include "Tasks.h"
```
manthanand, line 7: I think you can remove this line. This file is already included in RTOS_BPS.h.

Champers5000:
I made some sketchy changes on the wrapper function to reduce the number of parameters. Someone needs to check if this would work logically.

Tasks/Inc/RTOS_BPS.h
```diff
@@ -64,11 +71,26 @@ void RTOS_BPS_MutexPend(void);
...
+ * @param   *p_args - pointer to task function arguments
+ * @param   *prio - task priority
+ * @param   *p_stk_base - the stack
+ * @param   stk_size - size of the stack
+ * @param   *p_err - return error code
+ * @return  nothing to see here
  */
-void RTOS_BPS_TaskCreate(void);
+void RTOS_BPS_TaskCreate(
+    OS_TCB              *p_tcb,
```
connorl309, line 86: typedef this to be BPS_OS_TCB for now

Tasks/Inc/RTOS_BPS.h
```diff
@@ -64,11 +71,26 @@ void RTOS_BPS_MutexPend(void);
...
+    BPS_OS_TASK_NAME    *p_name,
+    BPS_OS_TASK_PTR      p_task,
+    void                *p_arg,
+    BPS_OS_TASK_PRIO     prio,
+    CPU_STK             *p_stk_base,
```
connorl309, line 91: same here

Tasks/Src/RTOS_BPS.c
```diff
@@ -38,9 +40,33 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+void RTOS_BPS_TaskCreate(
+    OS_TCB        *p_tcb,
+    CPU_CHAR      *p_name,
+    OS_TASK_PTR    p_task,
+    void          *p_arg,
+    OS_PRIO        prio,
+    CPU_STK       *p_stk_base,
+    CPU_STK_SIZE   stk_size,
+    OS_ERR        *p_err)
+    {
```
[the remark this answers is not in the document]
connorl309: To add onto this @Champers5000 if you find the source code for Micrium's TaskCreate, you will see a lot of these don't really need special types. TASK_NAME is just a char pointer, TASK_PTR is just a function pointer, etc. 

This func seems hard to abstract out because of how specific Micrium requires these things. FreeRTOS is a bit better - https://www.freertos.org/a00125.html.
This function will be rewritten when we port to FreeRTOS, anyways.

Tasks/Src/RTOS_BPS.c
```diff
@@ -58,9 +58,49 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+    BPS_OS_TASK_NAME    *p_name,
+    BPS_OS_TASK_PTR      p_task,
+    void                *p_arg,
+    BPS_OS_TASK_PRIO     prio,
+    CPU_STK             *p_stk_base,
+    BPS_OS_TASK_STK_SIZE stk_size
+    )
+    {
+        BPS_OS_ERR err;
+        OS_TCB *tcbptr = calloc(1, sizeof(OS_TCB));
```
connorl309, line 85: Do not use any memory allocation functions. Micrium does not guarantee these will work at all, and usually doing dynamic memory allocations on embedded systems is a nightmare and makes things explode. There's got to be a different way to do this.

Tasks/Src/RTOS_BPS.c
```diff
@@ -58,9 +58,49 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+        CPU_STK *stkptr = calloc(stk_size, sizeof(unsigned int)); // using size of unstacked int because cpu stack is type defined as an array of unsigned int
+        OSTaskCreate(tcbptr, p_name, p_task, p_arg, prio, stkptr, WATERMARK_STACK_LIMIT, stk_size, 0, 10,(void *)0, OS_OPT_TASK_STK_CHK | OS_OPT_TASK_SAVE_FP, &err);
```
connorl309, line 87: Why use WATERMARK_STACK_LIMIT? Shouldn't you use the function parameter stk_size? We want to use all of the task stack we can.

Tasks/Src/RTOS_BPS.c
```diff
@@ -58,9 +58,49 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
+        assertOSError(err);
+    }
+
+/*
+void RTOS_BPS_TaskCreate(
```
connorl309, line 92: remove this commented block

connorl309 (changes requested):
This needs changes. You aren't using some of the function parameters you state will be used when calling BPS's TaskCreate(), such as the stack size, or provided tcb. Note my comments on your calloc use as well. You can remove those statements and just use the function parameters.

Tasks/Src/RTOS_BPS.c
Champers5000: ((# •`⏏´•))   No

Tasks/Src/RTOS_BPS.c
Champers5000: https://docs.silabs.com/micrium/latest/micrium-general-concepts/04-memory-segments-and-lib-heap

Instead of calloc/malloc would using the memory allocation functions provided by the OS work? 
This way I can still reduce the amount of parameters.

ray4477 (approved):
Looks good to me

Tasks/Src/RTOS_BPS.c
Champers5000: Nah, tis too much work. I'll just leave as is without the calloc calls

connorl309: You should just be using your function parameters, not allocating any memory yourself. Remove the *tcbptr = calloc and the *stkptr = calloc, and just use the tcbptr and stkptr you pass into your function.
`OS_TCB              *p_tcb,
    BPS_OS_TASK_NAME    *p_name,
    BPS_OS_TASK_PTR      p_task,
    void                *p_arg,
    BPS_OS_TASK_PRIO     prio,
    CPU_STK             *p_stk_base,
    BPS_OS_TASK_STK_SIZE stk_size`
    these ones

Tasks/Src/RTOS_BPS.c
```diff
@@ -83,19 +83,17 @@ void RTOS_BPS_MutexPost(BPS_OS_MUTEX* mutex, BPS_OS_OPT options) {
...
  */
 void RTOS_BPS_TaskCreate(
-    OS_TCB              *p_tcb,
-    BPS_OS_TASK_NAME    *p_name,
-    BPS_OS_TASK_PTR      p_task,
-    void                *p_arg,
-    BPS_OS_TASK_PRIO     prio,
-    CPU_STK             *p_stk_base,
-    BPS_OS_TASK_STK_SIZE stk_size
+    BPS_OS_TCB      *p_tcb,
```
connorl309, line 86: Nice yeah, these look much better. This should probably work but we'll need to start doing comprehensive tests for the wrapper functions eventually.

connorl309 (approved):

Tasks/Src/RTOS_BPS.c
manthanand: you should remove this comment block tho lol

manthanand:
Could you also replace all the function calls in the test files as well?

Tests/Test_Amperes_Task.c
```diff
@@ -59,67 +59,46 @@ void Task1(void *p_arg){
                 &err);
     assertOSError(err);
 
-    RTOS_BPS_MutexCreate(&WDog_Mutex, "Watchdog Mutex");
+    OSMutexCreate(&WDog_Mutex,
```
manthanand, line 62: Is there a reason this was changed to the older version?

Tests/taskcreatereplace.py
```diff
@@ -0,0 +1,53 @@
+import os
```
manthanand, line 1: Could you add comments to the top of this file just explaining what it does and how it can be modified to replace other things. I don't really wanna keep this in the repository, but it could be used by other people while we are making major code changes. Also rename it to be a bit more general.

Tests/Test_Amperes_Task.c
Champers5000: Oh I accidentlly messed up this file so I copy pasted it from main, my bad.

PR #484, opened 2022-09-25
Simulator v2

ClarkPoon:
New version of the simulator. This one was designed with the intention that it would be easier to maintain than the old one, easier to create, and designed for running unit tests. It was not designed to be a completely faithful simulation of the BPS. I made a conscious decision to trade off fidelity to get better development time and implementation simplicity. It runs on the version of Micrium that we used for the previous simulator. A quick overview of it is:

Simulating Output:
Whenever the BPS sends output to a peripheral, it will be simulated by logging a record to a log file. This log file can be parsed after the test is complete to determine if it was successful.

Simulating Input:
JSON files can be used to define test cases. At startup, the simulator will read the JSON file and create a linked list of states that it will use to return fake data to the BSP layer whenever the BPS tries to read inputs. Each state has an associated time and will transition to the next state when the current state's time has run out. After the last state has completed, the simulator will shut down. Right now, state transitions only happen when a call into the simulator is made. I chose to use JSON because it is a common file format and many languages have libraries for it, so it should be relatively straightforward to create scripts/programs for generating test cases. For example JSON files, see https://github.com/lhr-solar/BPS/tree/simulator_v2/BSP/Simulator_V2/Data

What to simulate:
In the previous simulator, the well-defined boundary between the simulator and the non-simulated code was that the BSP was part of the simulator and the Driver layer and above was (mainly) unmodified. This forced the simulator to maintain compatibility with the Driver layer, which was expensive (especially for the LTC6811). To solve this problem, the new simulator implements it's own version of the BSP layer, but it also is allowed to modify the layers above it. This can be done safely by using the preprocessor to check if `SIMULATION` is defined to decide which parts of the upper layers are used, which allows the simulator to bypass the LTC6811 driver adn deal with a higher layer of abstraction (voltage and temperature readings) instead of a more complex lower layer (LTC6811 configuration)

Compiling  the Simulator:
Right now the simulator can be compiled with `make simulator_v2`. The simulator requires an additional `INPUT=` command line argument to specific the path to the JSON file for the test case to run. I chose to define the test at compile time instead of run time because I wanted to avoid making changes to `main()` to add command line arguemnts when I started this, but I am open to revisiting this if we decide a run-time option would be better. The `TEST=` option is also supported for running custom test cases.

Running the Simulator:
The simulator can be run with `./bps-simulator.out`

Checking tests:
This is not implemented yet, but I would like to create a script/program for parsing the simulator output files to determine the ending state of the simulator's outputs. We can then compare the final state with the expected state (define by the creator of the test case) to determine if the test passed or failed. This would allow us to script the testing process and integrate it with GitHub actions to run regression tests on all of our pull requests.

Some things (like current, CAN, and automated test checking) are not supported yet, so maybe we want to wait a little bit before merging with master? I also haven't written any serious test cases yet.


ClarkPoon:
Example output file:
image

manthanand:
Could the description you wrote be added to a README in the simulatore_v2 file?
Could you also remove all the old simulator files?

Apps/Src/Temperature.c
```diff
@@ -28,9 +31,17 @@ static int32_t maxTemperature;
 // 0 if discharging 1 if charging
 static uint8_t ChargingState;
 
+// simulator bypasses ltc driver
+#ifndef SIMULATION
 // Interface to communicate with LTC6811 (Register values)
 // Temperature.c uses auxiliary registers to view ADC data and COM register for I2C with LTC1380 MUX
 static cell_asic *Minions;
```
manthanand, line 38: So would Minions not be used if the simulator was running? Could the simulator not write to certain locations within the Minions struct or is that because the implementation would get too complicated? I'm wondering since some applications return data straight from the struct.

Apps/Src/Temperature.c
```diff
@@ -51,6 +62,9 @@ static inline int32_t median(int32_t a, int32_t b, int32_t c) {
  * @param boards LTC6811 data structure that contains the values of each register
  */
 void Temperature_Init(cell_asic *boards){
+// simulator bypasses ltc driver
+#ifndef SIMULATION
```
manthanand, line 66: I'm not a fan of this implementation. If we ever want to use the LTC anywhere we have to surround it with these guards. I also believe that we could test more (for example, someone trying to access the LTC without mutex protection) if we just rewrote the LTC drivers for simulation (could literally just be return some values from the json file).

manthanand: Or you could do something where if simulation is defined, replace all the LTC functions with some simulator functions. Though that might get more complicated.

BSP/Simulator_V2/Src/BSP_Fans.c
```diff
@@ -0,0 +1,65 @@
+#include "BSP_Fans.h"
+#include "Simulator.h"
+
+#define NUM_FANS 4
+
+static bool initialized = false;
+static uint32_t speed[NUM_FANS] = {0,0,0,0};
```
manthanand, line 7: Just for style/management, shouldn't the initialization of the speed happen in BSP_Fans_Init? So you can change the `#define` for the NUM_FANS without having to change anything else in the file.

BSP/Simulator_V2/Src/BSP_Fans.c
```diff
@@ -0,0 +1,65 @@
...
+ErrorStatus BSP_Fans_SetAll(uint32_t speed) {
+    ErrorStatus result = SUCCESS;
+    for (uint8_t i = 1; i <= 4; i++){
+        ErrorStatus e = BSP_Fans_Set(i, TOPSPEED);
+        if (e != SUCCESS) {
+            result = e;
+        }
+    }
+    return result;
+}
```
manthanand, line 65: newline?

BSP/Simulator_V2/Src/BSP_Fans.c
```diff
@@ -0,0 +1,65 @@
+#include "BSP_Fans.h"
+#include "Simulator.h"
+
+#define NUM_FANS 4
```
manthanand, line 4: Should move this to config.h so we don't have to change for both BSP layers if we do change.

BSP/Simulator_V2/Src/BSP_SPI.c
```diff
@@ -0,0 +1,203 @@
+/* Copyright (c) 2022 UT Longhorn Racing Solar */
+
```
manthanand, line 2: Does this file need to be rewritten (so needs TODO's)?

BSP/Simulator_V2/Src/Simulator.c
```diff
@@ -0,0 +1,227 @@
...
+#define JSON_PATH "Error: make sure you define the simulator JSON path in the Makefile!"
+#else
+// #define QUOTE_MARK "
+#define STR(X) # X
+#define STRINGIFY(X) STR(X)
+#define JSON_PATH STRINGIFY(SIMULATOR_JSON_PATH) 
+#endif
+
+// log something to the simualtor's log file
+void Simulator_log(char *str) {
```
manthanand, line 44: Should this have parameters similar to python logging, where you log errors, info, or warnings? I feel it would be easier to tell if something broke by just checking for an error instead of making sure when people write log data, they have to write `Hard Fault: ...`

Apps/Src/Temperature.c
ClarkPoon: Supporting arbitrary reads from Minions would require building a model of the LTC6811 to make sure all of the registers in the Minions struct are correct

Apps/Src/Temperature.c
ClarkPoon: Yeah, we could simulate the LTC through the driver, but it would be a lot of work. The reason I did it this way here is because I think it's better to get something that works and doesn't simulate the LTC driver than delay getting a running simulator by a month or two to get the LTC driver properly simulated.

If you want, we can always go back later and upgrade the simulator by making a simulator of the LTC driver.

One concern I have with this though is that if we find bugs with how we are interfacing with the hardware (like the timing issues we ran into for temperature), the changes might break the simulator. So swapping in the fake values here will probably reduce simulator maintenance.

BSP/Simulator_V2/Src/BSP_Fans.c
ClarkPoon: Ok. It didn't complain when I compiled it though, so maybe that's just the arm-gcc compiler

BSP/Simulator_V2/Src/BSP_SPI.c
ClarkPoon: Yeah, this one is just a copy-paste from simulator v1 and needs to be updated

BSP/Simulator_V2/Src/Simulator.c
ClarkPoon: Yeah, I can change it to something like that

Makefile
```diff
@@ -16,12 +16,18 @@ export TEST
...
+INPUT = none
+export INPUT
+
 stm32f413:
 	$(MAKE) -C BSP -C STM32F413 -j
 
 simulator:
 	$(MAKE) -C BSP -C Simulator -j
 
+simulator_V2:
```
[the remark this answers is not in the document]
ClarkPoon: Yeah, I left it in for now in case it has some code that's useful for the new simulator

ClarkPoon:
> Could the description you wrote be added to a README in the simulatore_v2 file?
> 
> Could you also remove all the old simulator files?

I can update the README. I was waiting on deleting the old simulator files until I had all of the BSP done (in case the old simulator had code we wanted to us)

Apps/Src/Temperature.c
manthanand: What I imagined for an LTC simulator was pretty much having it copy data that we give it from the JSON file to the Minions Struct. I'm not sure if that is a full blow simulator or not, but if it isn't worth it then this is fine for now too.

ClarkPoon: We could do that, but I think it would be annoying to figure out the right stuff to populate the JSON with

BSP/Simulator_V2/Src/BSP_Fans.c
ClarkPoon: Will do

manthanand:
Why is there a BSP_Strobelight file in this PR? Shouldn't strobelight just be connected to fault pin?

PR #486, opened 2022-10-01
Pid loop

ray4477:
PID loop app created. Changes fan speed based on current and desired temperature. Also implemented in VoltTempMonitor.task. PID constants are gibberish for now. No test files changed. 
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 


Apps/Src/PID.c
```diff
@@ -0,0 +1,41 @@
...
+int32_t PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
+    Error = DesiredTemp - InputTemp;
+    ErrorSum = ErrorSum + Error;
+
+    if (PreviousError == 0) {PreviousError = Error;} //init previous val first time
+
+    Rate = Error - PreviousError;
+    PreviousError = Error;     //updates previous err value
+
+    if (((PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/25000) > 4000) {
```
connorl309, line 31: So far looks good but this divisor likely going to need to be tuned alongside the initial values.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -149,7 +149,7 @@ ErrorStatus BSP_CAN_Write(uint32_t id, uint8_t data[8], uint8_t length) {
     
     gTxMessage.StdId = id;
     gTxMessage.DLC = length;
-	for(int i = 0; i < length; i++){
+	for(int i = 0; i < length && i < 8; i++){
```
connorl309, line 152: Don't think you need the i < 8 here, as length should never exceed 8 on calls.

connorl309 (commented):
Overall pretty clean I think. Left a couple comments. Agree on needing the physical pack to do testing.

BSP/STM32F413/Src/BSP_CAN.c
ray4477: oh, this was Clark's edit. It works with GCC 11.3+ and is needed because the compiler is too big brain. Honestly should've removed it and included it in a sep branch/commit but it's relatively minor

ClarkPoon: Yeah, it doesn't compile if you don't have this or connor's fix. I think we might also be able to fix this by changing `uint8_t data` to a `uint8_t *` instead of a `uint8_t [8]`, but I haven't tested it.

ClarkPoon: Yeah, it needs this to compile

manthanand:
Remove all changes from RTOS_Porting stuff. This should only contain PID application changes.

tasniabari (approved):
Good job on coding style, add defines for any random numbers used once they can be experimentally determined via testing with the battery pack.

connorl309 (approved):
Overall looks good but why are there changes in the RTOS BPS code? Make sure these changes aren't pushed through I'm thinking it will have some problems

Apps/Inc/PID.h
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @brief Gives fan speed based on Average temperature of pack and past error values
+ * @param InputTemp - current temperature
+ * @param DesiredTemp - desired temperature
+ * @return FanSpeed: 0-4000 PWM
+ */
+int32_t PID_Output(int32_t InputTemp, int32_t DesiredTemp);
+
+#endif
```
manthanand, line 19: needs newline here

manthanand:
You have stuff from the porting branch in here. Get rid of that stuff since it is in another PR.

Apps/Inc/PID.h
manthanand, line 19: If this is going to just be used for temperature, then just include it in temperature.h. If it is going to be used for other PID later down the line, you should make it a more general function.

Config/Inc/config.h
```diff
@@ -69,6 +69,14 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 #define MAX_CHARGE_TEMPERATURE_LIMIT	45000	    // Max temperature limit (milliCelcius)	(recommended release: 45.00C)
 #endif
 
+#ifndef PID_DESIRED_TEMPERATURE
+#define PID_DESIRED_TEMPERATURE         38000       // Desired temperature   (milliCelcius) 
+#endif
+
+#ifndef PID_MAX_TEMPERATURE
+#define PID_MAX_TEMPERATURE             55000       // Max temperature where PID is overridden (milliCelsius)
```
manthanand, line 77: make this 45000

Apps/Inc/Temperature.h
```diff
@@ -124,4 +131,11 @@ ErrorStatus Temperature_SampleADC(uint8_t ADCMode);
  */
 int32_t Temperature_GetMaxTemperature(void);
 
+/**
+ * @brief Gives fan speed based on Average temperature of pack and past error values
+ * @param InputTemp - current temperature
+ * @param DesiredTemp - desired temperature
+ * @return FanSpeed: 0-4000 PWM
+ */
+int32_t PID_Output(int32_t InputTemp, int32_t DesiredTemp);
```
manthanand, line 140: Styles guide. Temperature Functions should have Temperature_ prefix.

Apps/Src/PID.c
```diff
@@ -0,0 +1,40 @@
...
+    PreviousError = Error;     //updates previous err value
+
+    if (((PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR) > MAX_FAN_SPEED) {
+        return MAX_FAN_SPEED;
+    }
+    if (((PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR) <= 0) {
+        return 0;
+    }
+    return (PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR;
+}
```
manthanand, line 40: Just include this in the temperature.c file if it is part of that library.

Tasks/Src/Task_VoltTempMonitor.c
```diff
@@ -112,8 +115,19 @@ void Task_VoltTempMonitor(void *p_arg) {
             assertOSError(err);
 
             temperatureHasBeenChecked = true;
+        } 
+        //PID loop - sets fan speed based on avg temperature and desired temperature
+        //overrides PID loop if above PID_MAX_TEMPERATURE or if it's FAULT
+        if (PIDStatus != SAFE || temperatureStatus != SAFE) {
```
manthanand, line 121: if temperatureStatus != SAFE, then this code will never run. It would just make more sense to do something like this at line 103.
`if (Temperature_General_Check(PID_MAX_TEMPERATURE) != SAFE) BSP_Fans_SetAll(TOPSPEED);`
and so on. Also, it is kind of weird to have both functions (Temperature_CheckStatus and the one above) do basically the same thing with a slight change. I would add a TODO saying to merge the 2 functions together later down the line. For now it is fine though.

connorl309 (approved):
This looks good besides the RTOS_Porting stuff not being fully merged in (easy fix). I trust that the PID functionality calculations work but I agree that we will need the actual hardware to calculate the constants

manthanand (approved):

PR #487, opened 2022-10-01
Sim v2 cjson rewrite

connorl309:
Rewrote the main Simulator.c file to use the more powerful (and readable) cJSON functions. Also added a nearly-empty .json test file to make sure the dummy values get filled properly if fields don't exist in the current simulator state.


connorl309:
Ok I don't know why this PR is tracking all the changes I've made not on this branch but the only relevant commits are the bottom two.

BSP/Simulator_V2/Src/Simulator.c
```diff
@@ -40,6 +42,25 @@ static time_t startTime;
...
 #endif
 
+static uint16_t DUMMY_VOLTAGES[NUM_BATTERY_MODULES] = {
+    42, 42, 42, 42, 42,
+    42, 42, 42, 42, 42,
+    42, 42, 42, 42, 42,
+    42, 42, 42, 42, 42,
+    42, 42, 42, 42, 42,
+    42, 42, 42, 42, 42,
+    42
```
manthanand, line 52: If the json file does not specify the value of the voltages, make them a value that doesn't immediately trip the BPS.

connorl309: Got it

ClarkPoon: I think we would want to trip the BPS though if the test tries to measure the voltages, but we haven't provided voltage data, right?

connorl309: I think @ClarkPoon said that these voltage values should be ok, but I modified the temperature values to be a safe value of 30C.

BSP/Simulator_V2/Src/Simulator.c
```diff
@@ -88,46 +113,90 @@ static void readInputFile(void) {
...
+        // temperatureArray should have 31*2 elements.
+        cJSON* temperatureArray = cJSON_GetObjectItem(state, "temperatures");
+        cJSON* current = cJSON_GetObjectItem(state, "current");
+        cJSON* charge = cJSON_GetObjectItem(state, "charge");
+        cJSON* canList = cJSON_GetObjectItem(state, "can");
+    
+        // Check which state fields aren't present in the current sim state JSON.
+        // Lots of nasty if/else's here, not much way around this.
+        if (!time) {
+            printf("Current simulator state does not have a specified time value. Using -1...\n");
```
ClarkPoon, line 135: Might want to make these use the simulator log, but I don't feel strongly about it. We can always change it later if these get annoying

ClarkPoon (commented):
Have you tested this yet? If so, let me know and I'll approve if it's working

BSP/Simulator_V2/Src/Simulator.c
connorl309: I thought about that but decided on printf because you don't immediately see the output of the sim without opening the log file, and thought printfs() were a little more informative in the moment of running the sim.

connorl309:
Fixed merge conflict

connorl309:
Yep, it works for me. Manthan asked me to post screenshots of outputs so
I’ll start doing that.

ClarkPoon (approved):

PR #491, opened 2022-10-02
Compile line config ability

connorl309:
# Description
This PR modifies the build approach for the BPS code, and makes it more clean and descriptive to the user. Additionally allows the build to specify config.h values from the command line, in the format of:
`make DEFINES="-D VARIABLE=VALUE VARIABLE=VALUE....`. The build can also output verbosely, by passing READABLE=false to the make command.

An example is shown below.
image

# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 

### DIFF
--- a/BSP/STM32F413/Makefile
+++ b/BSP/STM32F413/Makefile
@@ -4,6 +4,7 @@
 TARGET = bps-leader
 
 DEBUG = 1
+READABLE = false
 
 ######################################
 # building variables
@@ -186,15 +187,32 @@ OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(ASM_SOURCES:.s=.o)))
 vpath %.s $(sort $(dir $(ASM_SOURCES)))
 OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(MESSAGES:.c=.o))) # messages
 
-$(BUILD_DIR)/%.o: %.c Makefile | $(BUILD_DIR) 
+# The ifneq... @... endif modifies the make console output to not print every file name,
+# assuming make is run with READABLE= some nonzero number
+
+$(BUILD_DIR)/%.o: %.c Makefile | $(BUILD_DIR) c_compile_msg
+ifeq ($(READABLE), true)
+	@$(CC) -c $(CFLAGS) -Wa,-a,-ad,-alms=$(BUILD_DIR)/$(notdir $(<:.c=.lst)) $< -o $@
+else
 	$(CC) -c $(CFLAGS) -Wa,-a,-ad,-alms=$(BUILD_DIR)/$(notdir $(<:.c=.lst)) $< -o $@
+endif
 
-$(BUILD_DIR)/%.o: %.s Makefile | $(BUILD_DIR)
+$(BUILD_DIR)/%.o: %.s Makefile | $(BUILD_DIR) asm_compile_msg
+ifeq ($(READABLE), true)
+	@echo Building ASM object files...
+	@$(AS) -c $(CFLAGS) $< -o $@
+else
 	$(AS) -c $(CFLAGS) $< -o $@
+endif
 
-$(BUILD_DIR)/$(TARGET).elf: $(OBJECTS) Makefile
+$(BUILD_DIR)/$(TARGET).elf: $(OBJECTS) Makefile build_msg
+ifeq ($(READABLE), true)
+	@$(CC) $(OBJECTS) $(LDFLAGS) -o $@
+	@$(SZ) $@
+else
 	$(CC) $(OBJECTS) $(LDFLAGS) -o $@
 	$(SZ) $@
+endif
 
 $(BUILD_DIR)/%.hex: $(BUILD_DIR)/%.elf | $(BUILD_DIR)
 	$(HEX) $< $@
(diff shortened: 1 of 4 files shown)


BSP/STM32F413/Makefile
```diff
@@ -45,7 +45,8 @@ $(wildcard $(RTOS_DIR)/uC-Serial/Driver/ST/*.c)	\
 $(wildcard $(RTOS_DIR)/uC-Serial/Line/*.c)	\
 $(wildcard $(RTOS_DIR)/uC-Serial/OS/*.c)	\
 $(wildcard $(RTOS_DIR)/uC-Serial/OS/uCOS-III/*.c)	\
-$(wildcard $(RTOS_DIR)/uC-Serial/Source/*.c)	
+$(wildcard $(RTOS_DIR)/uC-Serial/Source/*.c)	\
+$(wildcard $(TOP_DIR)/Config/Src/*.c)
```
ClarkPoon, line 49: Maybe add a comment that you are taking advantage of this being last in C_SOURCES (and how) in case we ever edit this in the future

Config/Inc/config.h
```diff
@@ -15,7 +15,12 @@ typedef enum ErrorStatus_e {ERROR = 0, SUCCESS = !ERROR} ErrorStatus;
 typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE = 3} SafetyStatus;
 
 // NUCLEO or custom
+#ifdef CUSTOM_COMMAND_SIZE
+#define COMMAND_SIZE CUSTOM_COMMAND_SIZE
+#else
```
ClarkPoon, line 20: @manthanand it might be a good time to delete the unused macros in config.h lol

Config/Inc/config.h
```diff
@@ -36,24 +41,70 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
...
 #define CHARGE_DISABLE_VOLTAGE          4000        // Voltage to stop charging at
+#endif
 
 // make sure we don't enable charging if we're too close to the voltage limit
 #if MAX_VOLTAGE_LIMIT - 100 < CHARGE_DISABLE_VOLTAGE
 #error "Charging maximum voltage is too close to voltage trip limit!"
 #endif
 
+#ifdef CUSTOM_MAX_DISCHARGE_TEMP
+#define MAX_DISCHARGE_TEMPERATURE_LIMIT CUSTOM_MAX_DISCHARGE_TEMP
```
ClarkPoon, line 74: @manthanand some of these names are also a bit verbose. `MAX` and `LIMIT` seem redundant together, for example. Might be annoying to type out on the command line

Config/Src/config_msgs.c
```diff
@@ -0,0 +1,58 @@
...
+ * #ifdef COMPILER_DEFINE_HERE
+ * #pragma message ("XYZ WAS SPECIFIED BY COMPILER")
+ * #endif
+ * 
+ * config.h will also have similar formatting.
+ */
+
+#ifdef CUSTOM_COMMAND_SIZE
+#pragma message ("Custom command size set!")
+#endif
```
ClarkPoon, line 16: I think we want to print out the values of the macros as well (assuming it's not obvious from looking at the make output). The motivation for this is that there were a few times last year we spent time debugging "bugs" when we had the wrong configuration in config.h

ClarkPoon (commented):
Some of my comments might be beyond the scope of this PR

Config/Inc/config.h
manthanand: I agree. but my assumption is that the verbose command that has to be typed out is "CUSTOM_whatever". I think that parameter should be changed to something easier to deal with. It would also be nice if all parameters were changed too be smaller since some of it is redundant as well.

Config/Inc/config.h
manthanand: Lines 17-24 can be deleted.

Config/Inc/config.h
ClarkPoon: Good point. You could give them the same name with something like
```C
#ifndef MAX_DISCHARGE_TEMP
#define MAX_DISCHARGE_TEMP 60000
#endif
```

connorl309: Yeah, I'll change them to use same name

Config/Src/config_msgs.c
connorl309: I'll try, unsure if the preprocessor supports this explicitly but there's probably a way to do it

connorl309:
Ok, I believe it's working. I've got it printing values as close to the end of the GCC output as possible - right before linking output of all the .o files generated. This is a sample test I did with just the first parameter.
image

ray4477 (approved):
looks good

manthanand:
One other functionality is adding a -DEBUG option that enables or disables the DEBUG parameter in the MAKEFILE. Default should be 1(enabled).

connorl309:
@manthanand see most recent commit - output is almost exactly what I want, but unsure why the ASM message is being printed four times. Something to do with the last few chunks of the makefile but I can't make heads or tails of it right now.
image

connorl309:
Working. Didn't use the DEBUG flag as it seemed to be modifying compilation flags, but can change that as necessary.

image

BSP/STM32F413/Makefile
```diff
@@ -4,6 +4,7 @@
 TARGET = bps-leader
 
 DEBUG = 1
+READABLE = false
```
manthanand, line 7: have default value be true.

MichaelMohn (approved):
Makes sense to me.

ray4477 (dismissed):
cool

Champers5000 (dismissed):
Printing looks good

manthanand (approved):

PR #496, opened 2022-10-08
Sim v2 timer

ray4477:
Did not emulate BPS_Timer_GetRunFreq because it is used only internally

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 

### DIFF
--- a/BSP/Simulator/Src/BSP_Timer.c
+++ b/BSP/Simulator/Src/BSP_Timer.c
@@ -17,7 +17,7 @@ static clock_t t;
  */
 void BSP_Timer_Init(void) {
     initialized = true;
-    Simulator_log("Initialized Timer\n");
+    Simulator_Log(LOG_INFO,"Initialized Timer\n");
 }
 
 /**
@@ -28,10 +28,10 @@ void BSP_Timer_Init(void) {
 void BSP_Timer_Start(void) {
     if (initialized) {
         t = clock();
-        Simulator_log("Timer started\n");
+        Simulator_Log(LOG_INFO,"Timer started\n");
     }
     else {
-        Simulator_log("Hard Fault: Initialize Timer before start\n");
+        Simulator_Log(LOG_INFO,"Hard Fault: Initialize Timer before start\n");
     }
 }
 
@@ -41,8 +41,11 @@ void BSP_Timer_Start(void) {
  * @return  Number of ticks
  */
 uint32_t BSP_Timer_GetTicksElapsed(void) {
-    return clock() - t;
-    
+    uint32_t TicksElasped = clock() - t;
+    char TimeMessage[100] = {0};
+    sprintf(TimeMessage, "Ticks Elasped: %d ticks\n", TicksElasped);
+    Simulator_Log(LOG_INFO, TimeMessage);
+    return TicksElasped;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


BSP/Simulator_V2/Src/BSP_Timer.c
```diff
@@ -50,6 +59,6 @@ uint32_t BSP_Timer_GetRunFreq(void) {
  * @return  Microseconds 
  */
 uint32_t BSP_Timer_GetMicrosElapsed(void) {
-    // TODO
-    return 42;
+    return (uint32_t) (BSP_Timer_GetTicksElasped() * 1000000) / CLOCKS_PER_SEC;
```
connorl309, line 62: is uint32 big enough for this or should we bump it up to u64

connorl309 (approved):

BSP/Simulator_V2/Src/BSP_Timer.c
ray4477: I think it should be big enough its like 4 billion for uint32. also Im not sure if i want to change to uint64 because functions should be the same as actual bsp layer

BSP/Simulator_V2/Src/BSP_Timer.c
```diff
@@ -50,6 +59,6 @@ uint32_t BSP_Timer_GetRunFreq(void) {
  * @return  Microseconds 
  */
 uint32_t BSP_Timer_GetMicrosElapsed(void) {
-    // TODO
-    return 42;
+    uint32_t MicrosElasped = (BSP_Timer_GetTicksElasped() * 1000000) / CLOCKS_PER_SEC;
```
tasniabari: This is kind of nitpicky and won't make a difference here, but it might be better to call BSP_Timer_GetRunFreq instead of CLOCKS_PER_SECOND for consistency with the function in the BSP layer.

tasniabari (approved):
Good programming style, logic seems sound.

BSP/Simulator_V2/Src/BSP_Timer.c
```diff
@@ -1,18 +1,22 @@
...
 #include "BSP_Timer.h"
+#include <time.h>
+#include "Simulator.h"
 #include <stdint.h>
 #include <stdlib.h>
 
 
-
+static bool initialized = false;
+static clock_t t;
```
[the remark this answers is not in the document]
ray4477: i used initialized vs Initialized because in other bsp simulator code like contactor and others it uses the lower case version

BSP/Simulator_V2/Src/BSP_Timer.c
ray4477: True but BSP_Timer_GetRunFreq is only called in this file nowhere else making no difference when used in simulator. Also implementing getrunfreq is i don't think possible or at least not worth the effort as its only use is in the getmicroselapsed.

manthanand:
Could you write a test for this feature?

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -28,18 +41,17 @@ void BSP_Timer_Start(void) {
...
- * @brief   Gets the running frequency of the timer (time per tick)
+ * @brief   Gets the ticks/s from CLOCKS_PER_SEC
  * @param   None
- * @return  frequency in Hz
+ * @return  ticks/s
  */
 uint32_t BSP_Timer_GetRunFreq(void) {
-    // TODO
-    return 42;
+    return CLOCKS_PER_SEC;  //returns ticks/s instead because this function is only used internally
```
manthanand, line 54: Leave a `NOTE:` above the function call that explains why this function wasn't implemented. Also, just return 0. It is easier to tell it isn't used.

connorl309 (approved):
Looks good. Seems pretty simple to implement. Make sure EOF has newline.

manthanand (approved):

PR #500, opened 2022-10-22
OSSemCreate wrapper

connorl309:
# Overview
Wrapper for the OSSemCreate function since we never had it for some reason

# Quality Assurance Checklist 

## Requirements 
- [ ] Followed Coding Style Guide 
- [ ] Code Build checks pass 
- [ ] No merge conflicts 
- [ ] Did you Have Fun?

## Things to Consider 


connorl309:
Ok the build is exploding for some reason. Not sure what's going on, looking into it.

tasniabari (approved):
No issues

Tasks/Inc/RTOS_BPS.h
```diff
@@ -103,6 +103,15 @@ void RTOS_BPS_DelaySecs(uint16_t dly);
...
 void RTOS_BPS_DelayMs(uint16_t dly);
 
+/**
+ * @brief Creates a semaphore with the initially specified count
+ * 
+ * @param sem - pointer to a semaphore object to create and initialize
+ * @param name - name of the semaphore
+ * @param count - initial count for the semaphore
+ */
+void RTOS_BPS_SemCreate(BPS_OS_SEM* sem, char* name, uint32_t count);
```
manthanand, line 113: FreeRTOS offers the option to add a max count that the sempahore can reach (before throwing an error or something). I would add something to the documentation saying it would be useful to add a certain constant in case we are improperly using a sempahore (when we do switch over). This happened with our CAN semaphore where it went up to the 100's because we were never posting it.

manthanand:
This PR is not complete. You need to replace all the function calls in all files.

connorl309:
> This PR is not complete. You need to replace all the function calls in all files.

Got it, will do. Was goign to do that once everything was merged into RTOS_Porting but doing it here makes more sense

ray4477 (commented):
I think it looks good. The only thing is the only time a sem create reference has a nonzero count is the semcreate for CANbus_MailSem4. this maybe a dumb question but can we set the initial resource count for that to zero? so we can just set all counts to zero in the wrapper?

manthanand:
No we cant. The reason for that semaphore is to check the number of CAN Mailboxes that are open. We start off with 3 open so we initialize it to 3.

PR #517, opened 2022-12-04
Fix Simulate Temperature where it reads index >= 61

ray4477:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 

### DIFF
--- a/Apps/Src/Temperature.c
+++ b/Apps/Src/Temperature.c
@@ -230,7 +230,7 @@ ErrorStatus Temperature_UpdateSingleChannel(uint8_t channel){
 #ifndef SIMULATION
 		rawTemperatures[board][channel][medianFilterIdx] = milliVoltToCelsius(Minions[board].aux.a_codes[0] / 10);
 #else
-		if (board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel < 61) {
+		if (board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel < 62) {
 			rawTemperatures[board][channel][medianFilterIdx] = Simulator_getTemperature(board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel);
 		}
 #endif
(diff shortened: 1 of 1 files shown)


connorl309 (approved):
looks good

Apps/Src/Temperature.c
```diff
@@ -230,7 +230,9 @@ ErrorStatus Temperature_UpdateSingleChannel(uint8_t channel){
 #ifndef SIMULATION
 		rawTemperatures[board][channel][medianFilterIdx] = milliVoltToCelsius(Minions[board].aux.a_codes[0] / 10);
 #else
-		rawTemperatures[board][channel][medianFilterIdx] = Simulator_getTemperature(board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel);
+		if (board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel < 62) {
```
ClarkPoon, line 233: Can you make it use the defines in config.h instead of a hard-coded 62?

Apps/Src/Temperature.c
```diff
@@ -230,7 +230,9 @@ ErrorStatus Temperature_UpdateSingleChannel(uint8_t channel){
 #ifndef SIMULATION
 		rawTemperatures[board][channel][medianFilterIdx] = milliVoltToCelsius(Minions[board].aux.a_codes[0] / 10);
 #else
-		rawTemperatures[board][channel][medianFilterIdx] = Simulator_getTemperature(board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel);
+		if (board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel < NUM_TEMPERATURE_SENSORS) {
+			rawTemperatures[board][channel][medianFilterIdx] = Simulator_getTemperature(board * MAX_TEMP_SENSORS_PER_MINION_BOARD + channel);
+		}
 #endif
```
manthanand, line 236: probably want to return error with an else

ray4477: the above statement `rawTemperatures[board][channel][medianFilterIdx] = milliVoltToCelsius(Minions[board].aux.a_codes[0] / 10);` i thought just filled all the temps up 64 regardless, so it would be an array {temps, repeated 62 times, 0, 0} or something. Also I think our simulator does tell user if we don't have 62 temperature values in our JSON file. if i did an error here though would it be to go into the fault state?

ClarkPoon: We should probably fix the bug in the non-simulator code, since we don't check the array bounds there either

manthanand (approved):

PR #522, opened 2023-01-09
Simulator bugfixes

manthanand:
# Quality Assurance Checklist 

## Requirements 
- [ ] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 

This PR fixes bugs in the simulator as well as bugs found by the simulator.


manthanand:
The build does not pass because simulate current is not properly implemented. However, some major bugs were fixed so I would still like to merge this in if it is approved.

connorl309 (approved):
Changes look good. sprintf modifications look like they will work a lot better for stack issues. There's a bug with the test verifier on Github (afaik it's infinitely running) but that's not really a BPS issue I don't think.

manthanand:
> Changes look good. sprintf modifications look like they will work a lot better for stack issues. There's a bug with the test verifier on Github (afaik it's infinitely running) but that's not really a BPS issue I don't think.

no its a simulator issue. Going to debug in this PR.

connorl309:
> > Changes look good. sprintf modifications look like they will work a lot better for stack issues. There's a bug with the test verifier on Github (afaik it's infinitely running) but that's not really a BPS issue I don't think.
> 
> no its a simulator issue. Going to debug in this PR.

Gotcha lmk if you get a lead on it and I'll try and help out
I cancelled the test bc it had been running for like 2 hours

BSP/Simulator/Src/BSP_SPI.c
```diff
@@ -6,18 +6,20 @@
...
+static bool initialized = false;
 
 /**
  * @brief   Initializes the SPI port.
  * @param   port The SPI port to initialize.
  * @return  None
  */
 void BSP_SPI_Init(spi_port_t port, bsp_os_t *spi_os, bool baremetal) {
-    // TODO
+    initialized = true;
```
ClarkPoon, line 20: Do we actually use SPI in the simulator? I thought we bypassed it everywhere

Drivers/Src/LTC2315.c
```diff
@@ -43,8 +46,10 @@ static void LTC2315_wakeup_sleep()
  */
 void LTC2315_Init(bsp_os_t spi_os) {
     BSP_SPI_Init(spi_ltc2315, &spi_os, true);
-
     LTC2315_wakeup_sleep();
+    #ifdef SIMULATION
+        
+    #endif
```
ClarkPoon, line 52: Why?

Drivers/Src/LTC2315.c
```diff
@@ -4,6 +4,9 @@
 #include "Tasks.h"
 #include "BSP_PLL.h"
 #include <stdio.h>
+#ifdef SIMULATION
+#include "Simulator.h"
+#endif
```
ClarkPoon, line 9: Why?

Tasks/Src/Task_FaultState.c
```diff
@@ -145,6 +142,8 @@ void EnterFaultState() {
 #ifdef DEBUGMODE
     char command[COMMAND_SIZE];
 #endif
+    BSP_WDTimer_Init(); //This is in case we did not pass critical state and watchdog timer was not initialized
+    BSP_WDTimer_Start();
```
ClarkPoon, line 146: Maybe we just want to not use the watchdog timer if this is the case because we don't want it to kick us out of the fault state?

ClarkPoon (commented):
We also still need to fix the race in Simulator_Transition, right?

manthanand:
> We also still need to fix the race in Simulator_Transition, right?

Im going to add an issue to fix it since base functionality works right now.

BSP/Simulator/Data/footest-out.json
```diff
@@ -1,18 +1,23 @@
 {
     "forbidden_states": {
-        "Contactor": "enabled"
+        "Contactor": [
+            ["ContactorC1", "enabled"],
+            ["ContactorCFAN", "enabled"]
```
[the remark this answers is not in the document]
manthanand: It is the contactor pin that is going through the Fan board.

BSP/Simulator/Src/BSP_SPI.c
manthanand: I just added some basic stuff since I was debugging it when debugging the LTC2315. Not really necessary, but doesn't hurt in case we do stuff with it later.

Drivers/Src/LTC2315.c
manthanand: idk man

BSP/Simulator/Data/footest-out.json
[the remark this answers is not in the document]
ClarkPoon: Can we give the contactors names based on what they are used for? Like array and main or something like that

Tasks/Src/Task_FaultState.c
manthanand: This is for actual BPS, not just the simulator. If we do not pass the critical state, then the fault state breaks because it tries to reset a timer that wasn't initialized. But if we do pass the critical state and then fault, then we need to reset the watchdog timer.

BSP/Simulator/Src/BSP_SPI.c
ClarkPoon: Ok. We just need to be aware that not all of the SPI traffic that there would be in the real system will show up in the simulator

Apps/Src/Contactor.c
```diff
@@ -9,9 +9,10 @@
  */
 void Contactor_Init(void) {
 	//Contactor Init is called before fan init, so initialize pwm here first. This will setup the contactor outputs
+	#ifdef SIMULATION
```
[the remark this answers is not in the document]
manthanand: Only ones I see are in fault state. I will undo it.

Tasks/Src/Task_FaultState.c
ClarkPoon: Yes. I'm suggesting that actual BPS should only try and reset the watchdog timer if the timer has been initialized already, since there is no point in running the watchdog timer if we are already in a fault state

BSP/Simulator/Data/footest-out.json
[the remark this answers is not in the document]
manthanand: C1 and CFAN are the GPIO Pins. On the upper layer, it is defined as such. 
image
I can still change it, but if we change a GPIO pin to mean another contactor, we would have to change it everywhere again.

BSP/Simulator/Src/BSP_SPI.c
manthanand: I will add a comment that says that.

BSP/Simulator/Data/footest-out.json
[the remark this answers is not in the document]
ClarkPoon: Oh, I see. That's fine then

Tasks/Src/Task_FaultState.c
manthanand: I see, I will add an issue ticket for that.

BSP/Simulator/Src/BSP_SPI.c
```diff
@@ -29,7 +32,24 @@ void BSP_SPI_Init(spi_port_t port, bsp_os_t *spi_os, bool baremetal) {
...
 void BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen) {
-    // TODO
+    char* spi = port ? "SPI3 LTC2315" : "SPI LTC6811";
+    char* msg;
+    if(!initialized) {
+        asprintf(&msg, "Used %s without initialization!\n", spi);
+        Simulator_Log_Location(LOG_ERROR, msg);
+        free(msg);
+        Fault_BitMap |= Fault_CRC;
+        EnterFaultState();
```
connorl309, line 42: I think officially we are supposed to post the Fault semaphore but for sim i dont think it really matters

i did the same last sem and it worked for the 1 test case we were debugging then

BSP/Simulator/Src/Simulator.c
```diff
@@ -180,18 +178,20 @@ static void readInputFile(char *jsonPath) {
                 tail->temperatures[idx] = cJSON_GetArrayItem(temperatureArray, idx)->valueint;
             }
         }
-        if (!current) { // current in mA
+        if (!currentObj) { // current in mA
             printf("Current simulator state does not have a specified current value. Using 30...\n");
             tail->current = 30;
+        } else {
+            tail->current = currentObj->valueint;
```
connorl309, line 185: this is the fix for current and charge not being set from the json (I forgot to set their values when rewriting the cjson parse stuff)

connorl309 (approved):
i think these changes all look good. probably some more bugs to figure out once we start testing on hardware but most of the sim bugs i was aware of seem fix(ed) in this pr

PR #533, opened 2023-02-10
Detect and fault on contactor fail, Merge Crit State into Check Contactors

pramukap:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [ ] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 


Tasks/Src/Task_Init.c
```diff
@@ -115,6 +116,16 @@ void Task_Init(void *p_arg) {
             TASK_IDLE_STACK_SIZE);
     
     CAN_Queue_Init();
+
+    // If a contactor is on before we turn it on in critical state, it may have failed and welded shut
+    if(Contactor_GetState(HVHIGH_CONTACTOR) || Contactor_GetState(HVLOW_CONTACTOR)) {
```
manthanand, line 121: This should not be in the init task. It should be in the CheckContactor task since that is the task that checks the contactor states.

manthanand:
Also, add some test to the simulator to test this functionality. I would rather not merge stuff in if it hasn't been tested.

Tasks/Src/Task_Init.c
ray4477: shouldnt it be in `task_critical `instead of `checkcontactor `because `checkcontactor `is initialized once the contactors are closed and `task_checkcontactors` is created

ray4477: * `check_contactor_task` is made in task critical

connorl309: Aren't we getting rid of the critical task as well, like we did with the FaultState task?

Tasks/Src/Task_Init.c
```diff
@@ -115,6 +116,16 @@ void Task_Init(void *p_arg) {
...
+        Fault_BitMap |= Fault_ESTOP;
+        RTOS_BPS_SemPost(&Fault_Sem4, OS_OPT_POST_1);
```
connorl309, line 123: Change this RTOS call to just EnterFaultState() or pull from master if you haven't

Tasks/Src/Task_Init.c
manthanand: yes

Tasks/Src/Task_Init.c
```diff
@@ -23,14 +24,23 @@ void Task_Init(void *p_arg) {
...
+                );
     //2
-    RTOS_BPS_TaskCreate(&CriticalState_TCB,	    // TCB
-            "TASK_CRITICAL_STATE",	            // Task Name (String)
-            Task_CriticalState,				    // Task function pointer
-            (void *)0,				            // Task function args
-            TASK_CRITICAL_STATE_PRIO,		    // Priority
-            CriticalState_Stk,				    // Stack
-            TASK_CRITICAL_STATE_STACK_SIZE);
+    // RTOS_BPS_TaskCreate(&CriticalState_TCB,	    // TCB
```
Champers5000, line 37: Why is the critical state replaced by checkcontactor task? Shouldn't we keep both

manthanand:
In this PR, I also want unused code to be deleted. Any mention of the Critical State Task should be removed. I would also like to see documentation updated in the BPS readme.

Docs/source/Software/Tasks.rst
```diff
@@ -182,7 +182,7 @@ Functionality:
     controls minion board that is hardwired to turn on when the BPS enters a fault state. For this purpose and due to 
     difficulty in implementing CAN with disabled interrupts, step 6 is skipped.
     
-    7) The WatchDog timer is continually reset to prevent the BPS from going into fault again.
+    7) The WatcsDog timer is continually reset to prevent the BPS from going into fault again.
```
tiandahuang, line 185: spelling

Tasks/Src/Task_CheckContactor.c
```diff
@@ -1,16 +1,56 @@
...
+        Fault_BitMap |= Fault_ESTOP;
+        EnterFaultState();
+    } else {
+        RTOS_BPS_SemPost(&SafetyCheck_Sem4, OS_OPT_POST_1);
+    }
+
+    // BLOCKING =====================
+    // Wait until voltage, open wire, temperature, contactors, and current(Amperes) are all checked and safe
+    for (int i = 0; i < NUM_FAULT_POINTS; i++){
+        RTOS_BPS_SemPend(&SafetyCheck_Sem4,OS_OPT_PEND_BLOCKING);
```
tiandahuang, line 29: these pends happen before the watchdog is launched, so a hang on initial safety checks won't be caught -- it'll be helpful to add a timeout on the ```SafetyCheck_Sem4``` checks and to fault on timeout.

Tasks/Src/Task_CheckContactor.c
```diff
@@ -1,16 +1,56 @@
...
-/* 
- * Launch this task once the contactor has been closed.
- * This will trip the BPS if the contactor is open.
- * The purpose of this is to trip the BPS if the emergency
- * cutoff switch is pressed.
- */
 void Task_CheckContactor(void *p_arg) {
     (void)p_arg;
+    
+    CANMSG_t CANMSG;
```
tiandahuang, line 13: do ```CANMSG_t CanMsg``` instead of ```CANMSG``` -- all caps reserved for macros and enums

tiandahuang: also you can do all this in one line, means you can get rid of the ```CanPayload``` variable:
```
CANMSG_t CanMsg = {.payload = {.idx = 0, .data.b = 1}};

File: Tasks/Src/Task_CheckContactor.c [hunk 7]
```diff
@@ -1,16 +1,56 @@
...
+    CANPayload_t CanPayload; 
+    CanPayload.idx    = 0;
+    CanPayload.data.b = 1;
+
+    // If a contactor is on before we turn it on in this task, it may have failed and welded shut
+    if(Contactor_GetState(HVHIGH_CONTACTOR) || Contactor_GetState(HVLOW_CONTACTOR)) {
```
tiandahuang, line 19: keep consistent spacing
```suggestion
    if (Contactor_GetState(HVHIGH_CONTACTOR) || Contactor_GetState(HVLOW_CONTACTOR)) {
```

Tasks/Src/Task_CheckContactor.c
```diff
@@ -1,16 +1,56 @@
...
+                );
+
+    // Turn Contactor On
+    Contactor_On(HVHIGH_CONTACTOR);
+    Contactor_On(ARRAY_CONTACTOR);
+    Contactor_On(HVLOW_CONTACTOR);
+
+    // Push All Clear message to CAN Q
+    CANMSG.id = ALL_CLEAR;
+    CANMSG.payload = CanPayload;
```
tiandahuang, line 49: because you're using the same ```CanPayload``` for both messages and just changing the ```CANMSG.id```, it would be clearer to move ```CANMSG.paylod = CanPayload;``` to the top during the declaration of ```CANMSG``` so it doesn't look like you forgot to set a payload for the ```CONTACTOR_STATE``` message

manthanand:
Since we got rid of the critical state task, you should also get rid of all mentions of the critical state task in the test files (including the ones that test the critical state task).

Tasks/Src/Task_CheckContactor.c
manthanand: We should update the function to pass in a timeout. Or you can create another function that is just for semaphore pending with timeout. Both would work, its just preference on coding style. In the first case, you would have to pass in NULL in 90% of cases where the function is used and update all the function calls. In the second case, you would have an almost duplicate function.

Tasks/Src/Task_CheckContactor.c
```diff
@@ -1,16 +1,52 @@
...
+    }
+
+    // BLOCKING =====================
+    // Wait until voltage, open wire, temperature, contactors, and current(Amperes) are all checked and safe
+    for (int i = 0; i < NUM_FAULT_POINTS; i++){
+        RTOS_BPS_SemPend(&SafetyCheck_Sem4,OS_OPT_PEND_BLOCKING);
+    }
+
+    // Turn Contactor On
+    Contactor_On(HVHIGH_CONTACTOR);
```
tiandahuang, line 40: more of a general question,
should we be doing contactors on then sending all clear on can or should we be sending all clear first?
idk if it really matters

after all we are 
just a speck of dust, spinning
in the universe

tiandahuang (approved):

Tasks/Src/Task_CheckContactor.c
```diff
@@ -1,16 +1,52 @@
 /* Copyright (c) 2018-2022 UT Longhorn Racing Solar */
-#include "RTOS_BPS.h"
+#include "config.h"
 #include "Contactor.h"
+#include "CANbus.h"
 #include "Tasks.h"
+#include "BSP_UART.h"
+#include "Voltage.h"
+#include "CAN_Queue.h"
```
manthanand, line 8: Why is UART & Voltage included in this?

Tasks/Src/Task_CheckContactor.c
manthanand: Lol, have to do all clear first because other systems would have to be powered by battery pack and not supplemental battery when they do their startup sequence.

Tasks/Src/Task_CheckContactor.c
```diff
@@ -1,16 +1,52 @@
...
+                TASK_PETWDOG_STACK_SIZE
+                );
+
+    // If a contactor is on before we turn it on in this task, it may have failed and welded shut
+    if (Contactor_GetState(HVHIGH_CONTACTOR) || Contactor_GetState(HVLOW_CONTACTOR)) {
+        Fault_BitMap |= Fault_ESTOP;
+        EnterFaultState();
+    } else {
+        RTOS_BPS_SemPost(&SafetyCheck_Sem4, OS_OPT_POST_1);
+    }
```
manthanand, line 31: What is the point of posting this semaphore if it is pended right after? I don't think it is necessary to do this and increase the value of NUM_FAULT_POINTS.

tiandahuang: to be fair num fault points would be a bit of a misnomer if it didn't include one of the fault points

manthanand: Lol, NUM_FAULT_POINTS includes open wire which we don't even check anymore. Ig it's fine to keep it. Just seems redundant imo.

Tasks/Src/Task_CheckContactor.c
manthanand: could u remove the unnecessary includes

manthanand (approved):

PR #548, opened 2023-02-18
Can fault state

MichaelMohn:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [ ] If applicable, does software feature have simulation test associated with it (.json & out.json file)
- [x] Did you Have Fun?

## Things to Consider 


connorl309:
Build issues need to be addressed but also make sure you git pull from master. There's some changes notably we are NOT using tasks for fault state and I believe going to migrate from using a task for critical state too.

BSP/STM32F413/STM32F4xx_StdPeriph_Driver/Inc/stm32f4xx_can.h
```diff
@@ -616,7 +616,8 @@ void CAN_TTComModeCmd(CAN_TypeDef* CANx, FunctionalState NewState);
 /* CAN Frames Transmission functions ******************************************/
 uint8_t CAN_Transmit(CAN_TypeDef* CANx, CanTxMsg* TxMessage);
 uint8_t CAN_TransmitStatus(CAN_TypeDef* CANx, uint8_t TransmitMailbox);
-void CAN_CancelTransmit(CAN_TypeDef* CANx, uint8_t Mailbox);
+void CAN_CancelTransCAN_Transmitmit(CAN_TypeDef* CANx, uint8_t Mailbox);
+bool findMailBox(CAN_TypeDef* CANx);
```
connorl309, line 620: what do we use this for

Tasks/Src/Task_FaultState.c
```diff
@@ -132,26 +132,58 @@ void EnterFaultState() {
...
 #endif
     BSP_WDTimer_Init(); //This is in case we did not pass critical state and watchdog timer was not initialized
     BSP_WDTimer_Start(); 
     while(1) {
+        //Send Trip Message
+        CANPayload_t payload;
+        payload.idx = 0;
+        payload.data.w = 1;
+        CANbus_BlockAndSend_FaultState(TRIP, payload);
+        delay_u(20000);
```
connorl309, line 155: is this enough/too little time?

Drivers/Src/CANbus.c
```diff
@@ -104,6 +115,53 @@ static ErrorStatus CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
...
+			txdata[0] = payload.idx;
+			memcpy(&txdata[1], &payload.data.w, sizeof(payload.data.w));
+			break;
+
+		// Handle invalid messages
+		default:
+			return ERROR;	// Do nothing if invalid
+	}
+
+	ErrorStatus retVal = BSP_CAN_Write(id, txdata, data_length);
```
tiandahuang, line 156: do while

Drivers/Src/CANbus.c
```diff
@@ -104,6 +115,53 @@ static ErrorStatus CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
 	return retVal;
 }
 
+static ErrorStatus CANbus_SendMsg_FaultState(CANId_t id, CANPayload_t payload) {
+	uint8_t txdata[8];
+	uint8_t data_length = 0;
+
+	// TODO: is it really best to keep the list of
```
tiandahuang, line 122: might as well take out the switch case; if we're in the fault state we don't really need this error checking because we can't handle it anyways

Drivers/Src/CANbus.c
```diff
@@ -121,6 +179,20 @@ ErrorStatus CANbus_BlockAndSend(CANId_t id, CANPayload_t payload) {
 	return result;
 }
 
+/**
+ * @brief   Transmits data onto the CANbus without mailbox semaphores.
+ * @param   id : CAN id of the message
+ * @param   payload : the data that will be sent.
+ * @return  ERROR if error, SUCCESS otherwise
+ */
+ErrorStatus CANbus_BlockAndSend_FaultState(CANId_t id, CANPayload_t payload) {
```
tiandahuang, line 188: this function seems kinda redundant but I guess we can keep it for interface consistency.
you should make this a one liner though, no need for the extra code and extra variables:
```return CANbus_SendMsg_FaultState(id, payload);```

File: Drivers/Src/CANbus.c [hunk 6]
```diff
@@ -39,18 +39,29 @@ static void CANbus_CountIncoming(void) {
...
+		RTOS_BPS_SemCreate(&CANbus_MailSem4, "CAN Mailbox Semaphore", 3); // # of mailboxes
+		RTOS_BPS_SemCreate(&CANbus_ReceiveSem4, "CAN Queue Counter Semaphore", 0);
+	}
 
-	RTOS_BPS_MutexCreate(&CANbus_TxMutex, "CAN TX Lock");
-	RTOS_BPS_MutexCreate(&CANbus_RxMutex, "CAN RX Lock");
-	RTOS_BPS_SemCreate(&CANbus_MailSem4, "CAN Mailbox Semaphore", 3); // # of mailboxes
-	RTOS_BPS_SemCreate(&CANbus_ReceiveSem4, "CAN Queue Counter Semaphore", 0);
 	// Initialize and pass interrupt hooks
     BSP_CAN_Init(CANbus_CountIncoming, CANbus_Release, loopback);
```
tiandahuang, line 54: I guess it's ok that ```CANbus_CountIncoming()``` and ```CANbus_Release()``` are RTOS-based functions for fault state since interrupts are disabled so this doesn't get called, but it would be nice to add a comment for clarification since it looks like it could be RTOS functions called during fault state

Alternatively pass in NULL if fault state to make things more clear

Apps/Src/FaultState.c
```diff
@@ -129,26 +142,66 @@ void EnterFaultState() {
...
     }
-    */
+    
 
 #ifdef DEBUGMODE
     char command[COMMAND_SIZE];
 #endif
     BSP_WDTimer_Init(); //This is in case we did not pass critical state and watchdog timer was not initialized
     BSP_WDTimer_Start(); 
+    int payload_id;
```
tiandahuang, line 159: ```payload_id``` should be named ```payload_idx``` (id and idx are very different things) -- also, should be a ```uint8_t``` since ```CANPayload_t``` idx is a ```uint8_t```

Apps/Src/FaultState.c
```diff
@@ -129,26 +142,66 @@ void EnterFaultState() {
...
+        //Send Current Readings
+        payload.idx = payload_id;
+        payload.data.w = Amps_GetReading();
+        CANbus_BlockAndSend_FaultState(CURRENT_DATA, payload);
+        payload_id++;
+        delay_u_copy(20000);
+        
+
+        //Send Voltage Readings
+        for (int i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module voltage data
```
tiandahuang, line 186: try to avoid using the ```int``` datatype -- you should be using ```uint8_t``` since all the functions you all have a ```uint8_t``` parameter and its bad to cast integer sizes down

Apps/Src/FaultState.c
```diff
@@ -12,13 +12,26 @@
...
-#ifndef SIMULATION
-#include "stm32f4xx.h"
-#else
+#include "BSP_PLL.h"
+#ifdef SIMULATION
 #include "Simulator.h"
 extern uint8_t stateCount;
 #endif
 
+void delay_u_copy(uint16_t micro)
```
tiandahuang, line 21: if you're going to copy over ```delay_u()```, make it static and make it more descriptive in naming (i.e. ```delay_u_faultstate()```)

tiandahuang: also a caution about the ```uint16_t``` parameter: you won't be able to have a delay higher than ~65.5ms

Apps/Src/FaultState.c
```diff
@@ -129,26 +142,66 @@ void EnterFaultState() {
...
+        CANPayload_t payload;
+        payload.idx = payload_id;
+        payload.data.w = 1;
+        CANbus_BlockAndSend_FaultState(TRIP, payload);
+        payload_id++;
+        delay_u_copy(20000);
+
+        // contactor message
+        payload.idx = payload_id;
+        payload.data.w = 2;
```
tiandahuang, line 172: TODO: find out what value to send

manthanand: Its in the sharepoint, this should be payload.data.b = 0

BSP/Inc/BSP_CAN.h
```diff
@@ -15,13 +15,13 @@
...
- * @brief   Transmits the data onto the CAN bus with the specified id
- * @param   id : Message of ID. Also indicates the priority of message. The lower the value, the higher the priority.
- * @param   data : data to be transmitted. The max is 8 bytes.
- * @param   length : num of bytes of data to be transmitted. This must be <= 8 bytes or else the rest of the message is dropped.
- * @return  ERROR if module was unable to transmit the data onto the CAN bus. SUCCESS indicates data was transmitted.
+ * @brief   Calls stm-level CAN_DeInit
+ * @return  None
  */
+void BSP_CAN_DeInit();
+
```
manthanand, line 24: You should have comments for both functions.

BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -28,6 +28,14 @@ void BSP_CAN_Init(callback_t rxEvent, callback_t txEnd, bool loopback) {
     Simulator_Log(LOG_INFO, "CAN Initialized\n");
 }
 
+/**
+ * @brief   Calls stm-level CAN_DeInit
+ * @return  None
+ */
+void BSP_CAN_DeInit() {
```
manthanand, line 35: You should also change the value of `CAN_Initialized` to false so more CAN messages aren't sent.

Drivers/Src/CANbus.c
manthanand, line 54: I'm confused, why do you pass in a faultstate variable into the driver level function for CAN initialization, and then use the tasks fault flag in the BPS level? Wouldn't it make more sense to do it the other way around (or the same way for both) so the layers don't cross? The BSP level should not be dependent on the tasks level IMO.

Drivers/Src/CANbus.c
manthanand: Wouldn't we need it for the memcpy changes?

Drivers/Src/CANbus.c
manthanand: wait yea, why don't you just make it one function instead of 2? I don't see a reason to have the other one be its own since nothing else will call it (and it is blocking).

Drivers/Src/CANbus.c
tiandahuang: https://github.com/lhr-solar/Embedded-Sharepoint

BSP/Inc/BSP_CAN.h
MichaelMohn: which functions? BSP_CAN_DeInit has them

Apps/Src/FaultState.c
```diff
@@ -12,13 +12,26 @@
...
 #include "Simulator.h"
 extern uint8_t stateCount;
 #endif
 
+void delay_u_copy(uint16_t micro)
+{
+  uint32_t delay = BSP_PLL_GetSystemClock() / 1000000;
+	for(uint32_t i = 0; i < micro; i++)
+	{
+		for(uint32_t j = 0; j < delay; j++);
```
Champers5000, line 26: Don't we need volatile ints for this

Apps/Src/FaultState.c
```diff
@@ -12,13 +12,26 @@
...
+#include "BSP_PLL.h"
+#ifdef SIMULATION
 #include "Simulator.h"
 extern uint8_t stateCount;
 #endif
 
+void delay_u_copy(uint16_t micro)
+{
+  uint32_t delay = BSP_PLL_GetSystemClock() / 1000000;
+	for(uint32_t i = 0; i < micro; i++)
```
connorl309, line 24: Does this need to be volatile, or will the compiler not optimize out the loop?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -114,29 +114,40 @@ void BSP_CAN_Init(callback_t rxEvent, callback_t txEnd, bool loopback) {
     gRxMessage.DLC = 0;
     gRxMessage.FMI = 0;
 
-    /* Enable FIFO 0 message pending Interrupt */
-    CAN_ITConfig(CAN1, CAN_IT_FMP0, ENABLE);
+    /* Enable interrupts if in normal state */
+    if(!Fault_Flag){
```
ray4477, line 118: It would be better if BSP_CAN_Init() has another parameter that is like `bool false` passed in as CANBUS_Init already has that bool than using a global flag for fault. Idk if using a global flag for fault would result in any race conditions if we initialize CAN before the fault flag could be set to 1

BSP/STM32F413/STM32F4xx_StdPeriph_Driver/Inc/stm32f4xx_can.h
MichaelMohn: outdated. this was fixed

Drivers/Src/CANbus.c
MichaelMohn: fixed

BSP/Simulator/Src/BSP_CAN.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Drivers/Src/CANbus.c
MichaelMohn: fixed

Drivers/Src/CANbus.c
MichaelMohn: fixed how tienda suggested. the function still exists but it is a one liner

Apps/Src/FaultState.c
```diff
@@ -129,26 +142,66 @@ void EnterFaultState() {
...
+            payload.idx = payload_id;
+            payload.data.w = voltage;
+            CANbus_BlockAndSend_FaultState(VOLT_DATA, payload);
+            payload_id++;
+            delay_u_copy(20000);
+        }
+
+        delay_u_copy(20000);
+        CANbus_BlockAndSend_FaultState(CURRENT_DATA, payload);
+        for (int i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module voltage data
```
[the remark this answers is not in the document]
MichaelMohn: I think i already am?
`        // contactor message
        payload.idx = payload_idx;
        payload.data.b = 0;
        CANbus_BlockAndSend_FaultState(CONTACTOR_STATE, payload);
        payload_idx++;
        delay_u_faultState(20000);`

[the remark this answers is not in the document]
MichaelMohn: i think i have this already

BSP/STM32F413/Src/BSP_CAN.c
MichaelMohn: fixed

Drivers/Src/CANbus.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Drivers/Src/CANbus.c
MichaelMohn: do nothing for now

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
```diff
@@ -129,26 +142,66 @@ void EnterFaultState() {
     // TODO: create an interrupt-independent CAN interface, so we can use CAN from within a fault state
     // avoid infinite recursive faults, since our CAN Driver relies on the OS to work
     // also don't call CAN if the watchdog tripped, since CAN won't be initialized
-    /*
+    
     if ((Fault_BitMap & (Fault_OS | Fault_WDOG)) == 0) {
```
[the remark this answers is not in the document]
MichaelMohn: fixed i think. can u check

Apps/Src/FaultState.c
```diff
@@ -12,13 +12,27 @@
...
 
+//Copy of delay_u function for Fault State
+static void delay_u_faultState(uint32_t micro)
+{
+  uint32_t delay = BSP_PLL_GetSystemClock() / 1000000;
+	for(volatile uint32_t i = 0; i < micro; i++)
+	{
+		for(volatile uint32_t j = 0; j < delay; j++);
+	}
+}
```
manthanand, line 29: this shouldn't be here. You should be using the BSP PLL function.

Apps/Src/FaultState.c
```diff
@@ -129,26 +143,64 @@ void EnterFaultState() {
...
+    //Reinit CAN in fault state
+    CANbus_Init(false, true);
+    
 
 #ifdef DEBUGMODE
     char command[COMMAND_SIZE];
 #endif
     BSP_WDTimer_Init(); //This is in case we did not pass critical state and watchdog timer was not initialized
     BSP_WDTimer_Start(); 
+    uint8_t payload_idx;
```
manthanand, line 158: why do you have this variable for payload idx? Couldn't you just set all the indexes with the value they should be?

Apps/Src/FaultState.c
```diff
@@ -129,26 +143,64 @@ void EnterFaultState() {
...
+        payload.data.w = Amps_GetReading();
+        CANbus_BlockAndSend_FaultState(CURRENT_DATA, payload);
+        payload_idx++;
+        delay_u_faultState(20000);
+        
+
+        //Send Voltage Readings
+        for (uint8_t i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module voltage data
+            int voltage = Voltage_GetModuleMillivoltage(i);
+            payload.idx = payload_idx;
```
manthanand, line 187: shouldn't this be `payload.idx = i`?

Apps/Src/FaultState.c
```diff
@@ -129,26 +143,64 @@ void EnterFaultState() {
...
+            payload.data.w = voltage;
```
manthanand, line 188: why have the variable voltage? just do `payload.data.w = Voltage_GetModuleMillivoltage(i)`

Apps/Src/FaultState.c
```diff
@@ -129,26 +143,64 @@ void EnterFaultState() {
...
+            CANbus_BlockAndSend_FaultState(VOLT_DATA, payload);
+            payload_idx++;
+            delay_u_faultState(20000);
+        }
+
+        delay_u_faultState(20000);
```
manthanand, line 194: do we need this delay?

Apps/Src/FaultState.c
```diff
@@ -129,26 +143,64 @@ void EnterFaultState() {
...
+    uint8_t payload_idx;
     while(1) {
+        payload_idx = 0;
+        //Send Trip Message
+        CANPayload_t payload;
+        payload.idx = payload_idx;
+        payload.data.w = 1;
+        CANbus_BlockAndSend_FaultState(TRIP, payload);
+        payload_idx++;
+        delay_u_faultState(20000);
```
manthanand, line 167: if the value in all the delays is the same, just make it a define at the top of the file so we can tune it easily when testing.

Apps/Src/FaultState.c
```diff
@@ -129,26 +143,64 @@ void EnterFaultState() {
...
+        }
+
+        delay_u_faultState(20000);
+        CANbus_BlockAndSend_FaultState(CURRENT_DATA, payload);
+        for (uint8_t i = 0; i < NUM_BATTERY_MODULES; i++){ //send all battery module temperature data
+            payload.idx = payload_idx;
+            payload.data.w = Temperature_GetModuleTemperature(i);
+            CANbus_BlockAndSend_FaultState(TEMP_DATA, payload);
+            payload_idx++;
+            delay_u_faultState(20000);
```
manthanand, line 201: same as voltage comments

BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -12,22 +12,36 @@
 #include <unistd.h>
 
 static bool CAN_Initialized = false;
+static bool FaultState = false;
```
manthanand, line 15: what is this variable for?

Drivers/Inc/CANbus.h
```diff
@@ -46,9 +46,16 @@ typedef struct {
 /**
  * @brief   Initializes the CAN system
  * @param   loopback	: if we should use loopback mode (for testing)	
+ * @param   faultState  : fault state determines whether to implement Rx and Tx interrupts
```
manthanand, line 49: should write what values mean what

Drivers/Src/CANbus.c
```diff
@@ -39,16 +39,29 @@ static void CANbus_CountIncoming(void) {
...
  */
-void CANbus_Init(bool loopback) {
-
-	RTOS_BPS_MutexCreate(&CANbus_TxMutex, "CAN TX Lock");
-	RTOS_BPS_MutexCreate(&CANbus_RxMutex, "CAN RX Lock");
-	RTOS_BPS_SemCreate(&CANbus_MailSem4, "CAN Mailbox Semaphore", 3); // # of mailboxes
-	RTOS_BPS_SemCreate(&CANbus_ReceiveSem4, "CAN Queue Counter Semaphore", 0);
-	// Initialize and pass interrupt hooks
-    BSP_CAN_Init(CANbus_CountIncoming, CANbus_Release, loopback);
+void CANbus_Init(bool loopback, bool faultState) {
```
manthanand, line 45: should add a note saying that you must call CANBus_DeInit() before CANbus_Init if you are reinitializing for Fault State.

Drivers/Src/CANbus.c
manthanand: Either remove the static function and place the logic in this function, or remove static from the above function. I see no point in having a function call a static function that isn't used in any other part of the canbus library unless you have a reason for this.

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: nope

Apps/Src/FaultState.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: fixed

Drivers/Inc/CANbus.h
MichaelMohn: fixed

Drivers/Src/CANbus.c
MichaelMohn: fixed

BSP/Simulator/Src/BSP_CAN.c
MichaelMohn: fixed

Apps/Src/FaultState.c
MichaelMohn: image
I'm not sure how to use this to set a delay

manthanand: You need to rebase with master. Raymond had a PR that added delay functions in the BSP_PLL library.

manthanand: If you look in master you will find them

Drivers/Src/CANbus.c
```diff
@@ -104,6 +117,52 @@ static ErrorStatus CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
...
+
+		// Handle invalid messages
+		default:
+			return ERROR;	// Do nothing if invalid
+	}
+
+	ErrorStatus retVal;
+	do{
+		retVal = BSP_CAN_Write(id, txdata, data_length);
+	} while(retVal != ERROR);
```
manthanand, line 161: don't we want to keep trying if retVal is ERROR? In this code, if it returns SUCCESS, it will keep resending the same message. Also, is there a point to having a return for this function (`retval`)? You don't actually use it in FaultState.c and you will never return an `ERROR` since you wait for `SUCCESS` to happen. If there isn't, then get rid of the variable and return altogether.

Drivers/Src/CANbus.c
```diff
@@ -39,16 +39,29 @@ static void CANbus_CountIncoming(void) {
 /**
  * @brief   Initializes the CAN system
  * @param   loopback	: if we should use loopback mode (for testing)	
+ * @param 	faultState  : determines whether to implement Rx and Tx interrupts
```
manthanand, line 42: fix the whitespace

BSP/Simulator/Src/BSP_CAN.c
manthanand: How was this fixed? I don't see the variable used anywhere except for it being set in `BSP_CAN_Init`?

Drivers/Src/CANbus.c
MichaelMohn: fixed

Drivers/Src/CANbus.c
MichaelMohn: i think this is just a github thing, I don't have that white space locally

Drivers/Src/CANbus.c
manthanand: You should use the `@NOTE` parameter. It's fine for now though

Drivers/Inc/CANbus.h
manthanand: I mean `false means interrupts are not enabled` or something like that

Drivers/Src/CANbus.c
```diff
@@ -104,6 +118,52 @@ static ErrorStatus CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
...
+
+		// Handle invalid messages
+		default:
+			return ERROR;	// Do nothing if invalid
+	}
+
+	ErrorStatus retVal;
+	do{
+		retVal = BSP_CAN_Write(id, txdata, data_length);
+	} while(retVal == ERROR);
```
manthanand, line 162: Lol, this doesn't have to be changed, but this could also be written as `while(BSP_CAN_Write(id, txdata, data_length) == ERROR)` since it would have to execute once. And we wouldn't need the extra variable.

Drivers/Src/CANbus.c
MichaelMohn: Removed Static

Drivers/Inc/CANbus.h
MichaelMohn: fixed

Drivers/Src/CANbus.c
```diff
@@ -104,6 +118,52 @@ static ErrorStatus CANbus_SendMsg(CANId_t id, CANPayload_t payload) {
...
+
+	return retVal;
```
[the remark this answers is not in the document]
MichaelMohn: fixed

manthanand:
The reason your whitespace looks wonky is because your indents are tabs and not spaces. That is not allowed per the coding styles guide because it makes the code look wonky on different machines. Please fix.

Drivers/Src/CANbus.c
manthanand: I don't think you understand what I'm saying. I don't think it is necessary to have a function call another function, so remove one of them. like delete it.

manthanand (approved):

Apps/Src/FaultState.c
```diff
@@ -129,26 +135,52 @@ void EnterFaultState() {
...
+
+        //Send Contactor Readings
+        payload.data.b = 0;
+        CANbus_SendMsg_FaultState(CONTACTOR_STATE, payload);
+        BSP_PLL_DelayU(MESSAGE_BUFFER);
+        
+        //Send Current Readings
+        payload.data.w = Amps_GetReading();
+        CANbus_SendMsg_FaultState(CURRENT_DATA, payload);
+        BSP_PLL_DelayU(MESSAGE_BUFFER);
```
ray4477, line 165: why is MESSAGE_BUFFER being used to delay micro seconds?
If MESSAGE_BUFFER is being used to delay it should be renamed, otherwise its confusing as it can be interpreted as the size of a message buffer

ray4477 (approved):

PR #561, opened 2023-02-25
update pid loop

MichaelMohn:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 


Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
+			ErrorSum = I_MAX_ACCUM;
+		}else{
+			ErrorSum = -I_MAX_ACCUM;
+		}
+	}
+
+	//5 is the estimated "hold output"
+	int32_t p_Output = ((-Error)/PROPORTION*1000) + 5; //Scale P-output to 0-8, Floor divide error by 1000
+
+	//I output could totally fuck things up (and probably will on the first test), so disable it and make sure p is good first
```
connorl309, line 446: remove the funny word from ur comment

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
+	//Only read error sum in range
+	if(abs(Error) < I_ZONE){
+		ErrorSum += Error;
+	}else{
+		ErrorSum = 0;
+	}
+
+	//Cap error sum at 500 read degrees
+	if(abs(ErrorSum) > I_MAX_ACCUM){
+		if(ErrorSum > 0){
```
connorl309, line 436: just do ErrorSum = abs(I_MAX_ACCUM)?

connorl309 (commented):
i dont know a whole lot about PID but a couple changes to make. overall structure looks good

Apps/Src/Temperature.c
ray4477: I don't think that would work because you're supposed to see if error goes over the max accum

Apps/Src/Temperature.c
ray4477: nah that can be left there fr

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
+
+	//5 is the estimated "hold output"
+	int32_t p_Output = ((-Error)/PROPORTION*1000) + 5; //Scale P-output to 0-8, Floor divide error by 1000
+
+	//I output could totally fuck things up (and probably will on the first test), so disable it and make sure p is good first
+	//Keep I gains low or you'll get weird oscillation. abs(I output) should not currently exceed 2
+	int32_t i_Output = ((-ErrorSum)/INTEGRAL*1000);
+
+	//Don't use D output
+	int32_t output = p_Output + i_Output;
```
ray4477, line 451: can be uint8_t

Apps/Src/Temperature.c
MichaelMohn: nah i did that because error sum needs to be able to be negative

Champers5000 (approved):
PID loop looks ok, has checks to ensure things don't go too horribly wrong

ray4477 (changes requested):
Change output to uint8_t if it only returns values from 0-8. I'm pretty sure your pid code works but instead of doing `(-ErrorSum)/INTEGRAL*1000` do `(-ErrorSum)*1000/INTEGRAL`. The otherway kind of gets rid of precison I'm p sure.

Apps/Src/Temperature.c
manthanand: lets keep it pg for now. maybe fudge

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
  * @brief Gives fan speed based on Average temperature of pack and past error values
  * @param InputTemp - current temperature
  * @param DesiredTemp - desired temperature
- * @return FanSpeed: 0-4000 PWM
+ * @return FanSpeed: 0-8
  */
 int32_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
```
manthanand, line 424: Why are you returning an int32 if the range of values is 0-8? Also, Idk what exactly your code does, and I'm kinda getting lost in the sauce with your algorithm (and the negative numbers). You should update the documentation for this

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
+	if(abs(ErrorSum) > I_MAX_ACCUM){
+		if(ErrorSum > 0){
+			ErrorSum = I_MAX_ACCUM;
+		}else{
+			ErrorSum = -I_MAX_ACCUM;
+		}
+	}
+
+	//5 is the estimated "hold output"
+	int32_t p_Output = ((-Error)/PROPORTION*1000) + 5; //Scale P-output to 0-8, Floor divide error by 1000
```
manthanand, line 444: should make `#define` for 1000 and 5 and write comments next to them saying what they are for. Same for the other defines that you added to the code.

Apps/Src/Temperature.c
```diff
@@ -43,16 +43,15 @@ static uint8_t currentChannel = 0;
...
-#define PROPORTION -1234
-#define INTEGRAL 4321
-#define DERIVATIVE -9876
-#define MAX_FAN_SPEED 4000
-#define DIVISOR 25000
+#define PROPORTION 2
+#define INTEGRAL 250
+#define I_ZONE 3000
+#define I_MAX_ACCUM 500000
+#define MAX_FAN_SPEED 8
```
manthanand, line 50: There is already a TOPSPEED define in Fans.h. You can either rename that to this or just use that. Having 2 defines for the same purpose is not a good idea.

Apps/Src/Temperature.c
tiandahuang: could replace all these if elses with ternaries

Apps/Src/Temperature.c
tiandahuang: ```static const uint32_t``` is a good alternative for local stuff, but these tunable parameters really should be in the header file

also name your ```#define``` more specific, i.e. ```TEMPERATURE_PID_KP``` or something, tbh make em long so we don't get any conflicts and so their purposes are very crystal clear

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
-    if (((PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR) > MAX_FAN_SPEED) {
-        return MAX_FAN_SPEED;
-    }
-    if (((PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR) <= 0) {
-        return 0;
-    }
-    return (PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR;
+	
+	//Only read error sum in range
+	if(abs(Error) < I_ZONE){
```
tiandahuang, line 428: I think ternaries are faster

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
+	//Don't use D output
+	int32_t output = p_Output + i_Output;
+
+	if(output > MAX_FAN_SPEED){
+		return MAX_FAN_SPEED;
+	}else if(output < 0){
+		return 0;
+	}else{
+		return output;
+	}
```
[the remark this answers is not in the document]
tiandahuang: it would help to use spaces between things like ```if (``` and ```} else {```

tiandahuang (changes requested):
please please read the style guide

Apps/Src/Temperature.c
MichaelMohn: fixed

Apps/Src/Temperature.c
MichaelMohn: fixed

MichaelMohn: fixed

MichaelMohn:
> Change output to uint8_t if it only returns values from 0-8. I'm pretty sure your pid code works but instead of doing `(-ErrorSum)/INTEGRAL*1000` do `(-ErrorSum)*1000/INTEGRAL`. The otherway kind of gets rid of precison I'm p sure.

Nah I might not understand this part of C perfectly... but at the moment ErrorSum is capped at 500,000 and integral is 250. So the max output is 2. If I place the 1000 in the numerator I would get a massively incorrect value right?

Apps/Src/Temperature.c
```diff
@@ -420,22 +419,42 @@ int32_t Temperature_GetMaxTemperature(void) {
...
+			ErrorSum = -I_MAX_ACCUM;
+		}
+	}
+
+	//5 is the estimated "hold output"
+	int32_t p_Output = ((-Error)/PROPORTION*1000) + 5; //Scale P-output to 0-8, Floor divide error by 1000
+
+	//I output could totally fuck things up (and probably will on the first test), so disable it and make sure p is good first
+	//Keep I gains low or you'll get weird oscillation. abs(I output) should not currently exceed 2
+	int32_t i_Output = ((-ErrorSum)/INTEGRAL*1000);
```
manthanand, line 448: For this code, if you are trying to multiply the fraction by 1000, then you need to have 1000 in the numerator. Otherwise, the first operation will round to 0 and you will always get 0 as the output (even if the value could be 1 or otherwise).

manthanand: If the denominator is to be multiplied by 1000, then change the parentheses to make it look like that.

manthanand: `-ErrorSum/(INTEGRAL*1000)`

MichaelMohn: fixed

Apps/Src/Temperature.c
MichaelMohn: fixed

Apps/Src/Temperature.c
MichaelMohn: fixed

Apps/Src/Temperature.c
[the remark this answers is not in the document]
MichaelMohn: fixed

Apps/Src/Temperature.c
MichaelMohn: fixed

manthanand: this isn't fixed

Apps/Src/Temperature.c
```diff
@@ -420,22 +413,26 @@ int32_t Temperature_GetMaxTemperature(void) {
...
 int32_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
     Error = DesiredTemp - InputTemp;
-    ErrorSum = ErrorSum + Error;
+	
+    //Only read error sum in range
+    ErrorSum += abs(Error) < TEMPERATURE_PID_I_ZONE ? Error : 0;
 
-    if (PreviousError == 0) {PreviousError = Error;} //init previous val first time
+	//Cap error sum at 500 read degrees
+    ErrorSum = abs(ErrorSum) > TEMPERATURE_PID_I_MAX_ACCUM ? ErrorSum > 0 ? TEMPERATURE_PID_I_MAX_ACCUM : -TEMPERATURE_PID_I_MAX_ACCUM : ErrorSum;
```
manthanand, line 425: add some parentheses here to make the ternaries easier to read. You can also write it in different lines so it is easier to understand.
```
ErrorSum = (abs(ErrorSum) > TEMPERATURE_PID_I_MAX_ACCUM) ? 
    ((ErrorSum > 0) ? TEMPERATURE_PID_I_MAX_ACCUM : -TEMPERATURE_PID_I_MAX_ACCUM) : ErrorSum;
```

Apps/Src/Temperature.c
```diff
@@ -420,22 +413,26 @@ int32_t Temperature_GetMaxTemperature(void) {
...
 
-    Rate = Error - PreviousError;
-    PreviousError = Error;     //updates previous err value
+	//5 is the estimated "hold output"
```
manthanand, line 427: you should still have the 5 as a define at the top of the file

Apps/Src/Temperature.c
```diff
@@ -420,22 +413,26 @@ int32_t Temperature_GetMaxTemperature(void) {
...
-        return 0;
-    }
-    return (PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR;
+	//I output could totally fudge things up (and probably will on the first test), so disable it and make sure p is good first
+	//Keep I gains low or you'll get weird oscillation. abs(I output) should not currently exceed 2
+	int32_t i_Output = ((-ErrorSum)/(TEMPERATURE_PID_INTEGRAL*TEMPERATURE_PID_MILICELCIUS_CONVERT));
+
+	//Don't use D output
+	int32_t output = p_Output + i_Output;
+    output = output > TOPSPEED ? TOPSPEED : output < 0 ? 0 : output;
```
manthanand, line 436: same as other comment with ternary

Apps/Src/Temperature.c
MichaelMohn: fixed

Apps/Src/Temperature.c
MichaelMohn: fixed

Apps/Src/Temperature.c
MichaelMohn: fixed

ray4477:
> > Change output to uint8_t if it only returns values from 0-8. I'm pretty sure your pid code works but instead of doing `(-ErrorSum)/INTEGRAL*1000` do `(-ErrorSum)*1000/INTEGRAL`. The otherway kind of gets rid of precison I'm p sure.
> 
> Nah I might not understand this part of C perfectly... but at the moment ErrorSum is capped at 500,000 and integral is 250. So the max output is 2. If I place the 1000 in the numerator I would get a massively incorrect value right?

It should be equivalent mathematical statements. The problem with the before is if ErrorSum is say less than 250 than if it is divided by INTEGRAL (250) it will become zero because int division rounds down. If you do the other way then it will return fine. Here is an example below
image

MichaelMohn:
> > > Change output to uint8_t if it only returns values from 0-8. I'm pretty sure your pid code works but instead of doing `(-ErrorSum)/INTEGRAL*1000` do `(-ErrorSum)*1000/INTEGRAL`. The otherway kind of gets rid of precison I'm p sure.
> > 
> > 
> > Nah I might not understand this part of C perfectly... but at the moment ErrorSum is capped at 500,000 and integral is 250. So the max output is 2. If I place the 1000 in the numerator I would get a massively incorrect value right?
> 
> It should be equivalent mathematical statements. The problem with the before is if ErrorSum is say less than 250 than if it is divided by INTEGRAL (250) it will become zero because int division rounds down. If you do the other way then it will return fine. Here is an example below image

This has been fixed I believe.

tiandahuang (approved):

Apps/Src/Temperature.c
```diff
@@ -418,24 +411,31 @@ int32_t Temperature_GetMaxTemperature(void) {
...
-    if (PreviousError == 0) {PreviousError = Error;} //init previous val first time
+    //Cap error sum at 500 read degrees
+    ErrorSum = (abs(ErrorSum) < TEMPERATURE_PID_I_MAX_ACCUM) ? ErrorSum :
+        (ErrorSum > 0 ? TEMPERATURE_PID_I_MAX_ACCUM : -TEMPERATURE_PID_I_MAX_ACCUM);
 
-    Rate = Error - PreviousError;
-    PreviousError = Error;     //updates previous err value
+    //5 is the estimated "hold output"
+    //Scale P-output to 0-8, Floor divide error by 1000
+    int8_t p_Output = (int8_t)((-Error)/(TEMPERATURE_PID_PROPORTIONAL*TEMPERATURE_PID_MILICELCIUS_CONVERT)) + TEMPERATURE_HOLD_FAN_SPEED;
```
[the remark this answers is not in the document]
tiandahuang: in addition you could add divide with rounding instead of divide with truncating:
https://stackoverflow.com/questions/2422712/rounding-integer-division-instead-of-truncating

Apps/Src/Temperature.c
```diff
@@ -418,24 +411,31 @@ int32_t Temperature_GetMaxTemperature(void) {
...
- * @return FanSpeed: 0-4000 PWM
+ * @return FanSpeed: 0-8
  */
-int32_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
+int8_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
     Error = DesiredTemp - InputTemp;
-    ErrorSum = ErrorSum + Error;
+	
+    //Only read error sum in range
+    ErrorSum += abs(Error) < TEMPERATURE_PID_I_ZONE ? Error : 0;
```
ray4477, line 422: why is this here, reading errors only in range of 3 C? Temperature_PID_I_ZONE is only 3000 which is 3 celsius. idk but feels low to me

Apps/Src/Temperature.c
```diff
@@ -418,24 +411,31 @@ int32_t Temperature_GetMaxTemperature(void) {
 
 /**
  * @brief Gives fan speed based on Average temperature of pack and past error values
- * @param InputTemp - current temperature
+ * @param InputTemp - current temperature. Must not exceed 284 or -284 Celcius given that expected temp is 38 celcius.
```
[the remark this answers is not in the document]
MichaelMohn: yeah i figured that if the car has reached 10 degrees below absolute zero that there would be more pressing issues at hand than the fans

Apps/Src/Temperature.c
MichaelMohn: 3 was a guess, i actually just increased it to 5. Regardless this number is very low because you don't want to use the "i-part" of the pid unless you are very close to the target. 

Error, instead of errorsum, is used by the "p-part" and is what is read and updated regardless of how close to the target we are.

Apps/Src/Temperature.c
```diff
@@ -418,24 +411,31 @@ int32_t Temperature_GetMaxTemperature(void) {
...
-        return 0;
-    }
-    return (PROPORTION*(Error) + INTEGRAL*(ErrorSum) + DERIVATIVE*(Rate))/DIVISOR;
+    //I output could totally fudge things up (and probably will on the first test), so disable it and make sure p is good first
+    //Keep I gains low or you'll get weird oscillation. abs(I output) should not currently exceed 2 
+    int8_t i_Output = (int8_t)((-ErrorSum)/(TEMPERATURE_PID_INTEGRAL*TEMPERATURE_PID_MILICELCIUS_CONVERT));
+    
+    //Don't use D output
+    int8_t output = p_Output + i_Output;
+    output = (output > TOPSPEED) ? TOPSPEED :
```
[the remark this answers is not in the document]
MichaelMohn: good idea

Apps/Src/Temperature.c
```diff
@@ -418,24 +411,31 @@ int32_t Temperature_GetMaxTemperature(void) {
...
 /**
  * @brief Gives fan speed based on Average temperature of pack and past error values
- * @param InputTemp - current temperature
+ * @param InputTemp - current temperature. Must not exceed 284 or -284 Celcius given that expected temp is 38 celcius.
  * @param DesiredTemp - desired temperature
- * @return FanSpeed: 0-4000 PWM
+ * @return FanSpeed: 0-8
  */
-int32_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
+int8_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp) {
```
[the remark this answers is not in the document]
MichaelMohn: oh yeah

Apps/Src/Temperature.c
MichaelMohn: jk going back to three

ray4477 (approved):

Apps/Src/Temperature.c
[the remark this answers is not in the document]
MichaelMohn: ok I'm dividing my millicelcius at the end for output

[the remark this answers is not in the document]
MichaelMohn: fixed

Apps/Src/Temperature.c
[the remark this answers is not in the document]
MichaelMohn: fixed

Apps/Src/Temperature.c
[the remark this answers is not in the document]
manthanand: you should provide error checking if we exceed these bounds. What happens if the code accidentally puts in those values when the actual temperature is fine or within bounds?

Apps/Inc/Temperature.h
```diff
@@ -130,5 +138,5 @@ int32_t Temperature_GetMaxTemperature(void);
  * @param DesiredTemp - desired temperature
  * @return FanSpeed: 0-4000 PWM
  */
-int32_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp);
+uint8_t Temperature_PID_Output(int32_t InputTemp, int32_t DesiredTemp);
```
manthanand, line 141: the inputs to this should also be all positive. Shouldn't they be uint32_t?

Apps/Src/Temperature.c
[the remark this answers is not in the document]
MichaelMohn: fixed

Apps/Inc/Temperature.h
MichaelMohn: fixed

Tests/Test_Fault_Lights.c
```diff
@@ -0,0 +1,67 @@
...
+    Fault_OTEMP   = 0x004,
+    Fault_OCURR   = 0x008,
+    Fault_OW      = 0x010,
+    Fault_HANDLER = 0x020,
+    Fault_OS      = 0x040,
+    Fault_WDOG    = 0x080,
+    Fault_CRC     = 0x100,
+    Fault_ESTOP   = 0x200,
+    Fault_Max
+}Fault_Set;
```
Champers5000, line 24: Wasn't this defined elsewhere already?

Champers5000 (dismissed):

Apps/Inc/Temperature.h
```diff
@@ -130,5 +139,5 @@ int32_t Temperature_GetMaxTemperature(void);
  * @param DesiredTemp - desired temperature
  * @return FanSpeed: 0-4000 PWM
```
manthanand, line 140: change comment here.

Apps/Src/Temperature.c
```diff
@@ -418,24 +411,34 @@ int32_t Temperature_GetMaxTemperature(void) {
 
 /**
  * @brief Gives fan speed based on Average temperature of pack and past error values
- * @param InputTemp - current temperature
+ * @param InputTemp - current temperature. Must not exceed 293 or -294 Celcius given that expected temp is 38 celcius.
```
manthanand, line 414: Write what happens if temperature does exceed these bounds.

Tests/Test_Fault_Lights.c
MichaelMohn: i didnt mean to push this file oops

Apps/Inc/Temperature.h
MichaelMohn: fixed

manthanand (approved):

PR #565, opened 2023-03-04
SPI Error Handling

Champers5000:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 

### DIFF
--- a/BSP/Inc/BSP_SPI.h
+++ b/BSP/Inc/BSP_SPI.h
@@ -47,7 +47,7 @@ void BSP_SPI_SetClock(spi_speed_t speed);
  * @param   txLen   length of data array.
  * @return  None
  */
-void BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
+bool BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
 
 /**
  * @brief   Gets the data from SPI. With the way the LTC6811 communication works, the LTC6811 will 
@@ -59,7 +59,7 @@ void BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
  * @param   rxLen   length of data array.
  * @return  None
  */
-void BSP_SPI_Read(spi_port_t port, uint8_t *rxBuf, uint32_t rxLen);
+bool BSP_SPI_Read(spi_port_t port, uint8_t *rxBuf, uint32_t rxLen);
 
 /**
  * @brief   Sets the state of the chip select output pin. Set the state to low/0 to notify the 
--- a/BSP/Inc/BSP_SPI.h
+++ b/BSP/Inc/BSP_SPI.h
@@ -57,7 +57,7 @@ bool BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
  * @param   port    the SPI port to read from
  * @param   rxBuf   data array to store the data that is received.
  * @param   rxLen   length of data array.
- * @return  None
+ * @return  false for fail, true for success
  */
 bool BSP_SPI_Read(spi_port_t port, uint8_t *rxBuf, uint32_t rxLen);
 
@@ -67,7 +67,7 @@ bool BSP_SPI_Read(spi_port_t port, uint8_t *rxBuf, uint32_t rxLen);
  *          make the LTC6811 go to standby.
  * @param   port    the SPI port to select/deselct on
  * @param   state   0 for select, 1 to deselect
- * @return  None
+ * @return  false for fail, true for success
  */
 void BSP_SPI_SetStateCS(spi_port_t port, uint8_t state);
 
--- a/BSP/Inc/BSP_SPI.h
+++ b/BSP/Inc/BSP_SPI.h
@@ -46,7 +46,7 @@ void BSP_SPI_SetClock(spi_speed_t speed);
  * @param   port    the SPI port to write to
  * @param   txBuf   data array that contains the data to be sent.
  * @param   txLen   length of data array.
- * @return  false for fail, true for success
+ * @return  ERROR on fail, SUCCESS on successful write
  */
 ErrorStatus BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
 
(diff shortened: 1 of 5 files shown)


BSP/Inc/BSP_SPI.h
```diff
@@ -47,7 +47,7 @@ void BSP_SPI_SetClock(spi_speed_t speed);
  * @param   txLen   length of data array.
  * @return  None
  */
-void BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
+bool BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
```
manthanand, line 50: Need to change the comments for the return value. Also, if this is an error, make it an `ErrorStatus` type.

BSP/Inc/BSP_SPI.h
```diff
@@ -57,17 +57,17 @@ void BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen);
...
 /**
  * @brief   Sets the state of the chip select output pin. Set the state to low/0 to notify the 
  *          LTC6811 that the data sent on the SPI lines are for it. Set the state to high/1 to 
  *          make the LTC6811 go to standby.
  * @param   port    the SPI port to select/deselct on
  * @param   state   0 for select, 1 to deselect
- * @return  None
+ * @return  false for fail, true for success
  */
 void BSP_SPI_SetStateCS(spi_port_t port, uint8_t state);
```
manthanand, line 72: This doesn't match it's comment.

BSP/Inc/BSP_SPI.h
manthanand: Replace all bool returns with `ErrorStatus`

manthanand (changes requested):

BSP/Inc/BSP_SPI.h
```diff
@@ -45,9 +46,9 @@ void BSP_SPI_SetClock(spi_speed_t speed);
  * @param   port    the SPI port to write to
  * @param   txBuf   data array that contains the data to be sent.
  * @param   txLen   length of data array.
- * @return  None
+ * @return  false for fail, true for success
```
manthanand, line 49: you did not update the comments.

BSP/Simulator/Src/BSP_SPI.c
```diff
@@ -40,6 +40,7 @@ void BSP_SPI_Write(spi_port_t port, uint8_t *txBuf, uint32_t txLen) {
         free(msg);
         Fault_BitMap |= Fault_CRC;
         EnterFaultState();
+        return ERROR;
```
manthanand, line 43: There in no point in this after `EnterFaultState()` since EnterFaultState never leaves itself. Also, the return value is based on whether a parameter is passed when the port is not one of the two defined. That functionality is not implemented in this c file.

Drivers/Src/LTC681x.c
```diff
@@ -46,21 +46,30 @@ Copyright 2017 Linear Technology Corp. (LTC)
...
   uint8_t data = 0;
   BSP_SPI_Read(spi_ltc6811, &data, 1);
 	return data;
 }
 
-static void spi_write_multi8(uint8_t *txBuf, uint32_t txSize){
-	BSP_SPI_Write(spi_ltc6811, txBuf, txSize);
+static ErrorStatus spi_write_multi8(uint8_t *txBuf, uint32_t txSize){
+  ErrorStatus status = BSP_SPI_Write(spi_ltc6811, txBuf, txSize);
+  if(status==false){
```
manthanand, line 59: This should not be false, it should be `ERROR`.

manthanand: Make these changes to the rest of the file as well.

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -87,15 +87,15 @@ static inline void SPI_WaitTx(spi_port_t port){
 /** SPI_WriteRead
  * @brief   Sends and receives a byte of data on the SPI line.
  * @param   txData single byte that will be sent to the device.
- * @return  rxData single byte that was read from the device.
+ * @return  rxData single byte that was read from the device. Returns 0xFF00 on error
```
tiandahuang, line 90: no magic numbers:
https://en.wikipedia.org/wiki/Magic_number_(programming)

give this return value a ```#define``` or (better) an ```enum```

BSP/STM32F413/Src/BSP_SPI.c
```diff
@@ -87,15 +87,15 @@ static inline void SPI_WaitTx(spi_port_t port){
...
-    if(port >= NUM_SPI_BUSSES) return -1;
-	
+static uint16_t SPI_WriteRead(spi_port_t port, uint8_t txData){
+    if(port >= NUM_SPI_BUSSES) return 0xFF00; //set first 8 bits to 1 to signal error
+
 	SPI_Wait(port);
 	SPI_BUSSES[port]->DR = txData & 0x00FF;
 	SPI_Wait(port);
-	return SPI_BUSSES[port]->DR & 0x00FF;
+	return SPI_BUSSES[port]->DR & 0x00FF; //normal return will have the first 8 bits as 0
```
tiandahuang, line 98: example implementation using enum instead of magic number:
```
enum spi_status_t 
        SPI_OK = 0,
        SPI_ERROR = -1 // or 0xFFFF
        SPI_other_status = 5 // or some other number

uint16_t return_status;
if error:
        return_status = SPI_ERROR
elif valid read but return status not ok:
        return_status = SPI_other_status
else // return status ok
        return_status = SPI_OK

return spi->data | return_status
```

manthanand: I think `#define` would work better since we do not have a different error handling scheme for SPI_ERROR vs SPI_other_status. If you did want to do an enum, then I would say something like this:
```
enum spi_status_t
    SPI_OK = 0x00FF
    SPI_ERROR = 0xFF00

...blah 

return spi->data && return_status
```

tiandahuang: ```spi->data && return_status``` would result in ```true``` most of the time hehe

tiandahuang: I'm a big fan of enums when you can use it though, because it has a defined type unlike ```#define```

manthanand (approved):

tiandahuang (approved):
lgtm 🚀

PR #575, opened 2023-03-25
Makefile print better

connorl309:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Did you Have Fun?

## Things to Consider 

### DIFF
--- a/BSP/STM32F413/Makefile
+++ b/BSP/STM32F413/Makefile
@@ -157,8 +157,9 @@ C_INCLUDES =  \
 # compile gcc flags
 ASFLAGS = $(MCU) $(AS_DEFS) $(AS_INCLUDES) $(OPT) -Wall -fdata-sections -ffunction-sections
 
-CFLAGS = $(MCU) $(C_DEFS) $(C_INCLUDES) $(OPT) -Wall -Werror -fdata-sections -ffunction-sections
-
+CFLAGS = $(MCU) $(C_DEFS) $(C_INCLUDES) $(OPT) -Wall -Werror -fdata-sections -ffunction-sections 
+CFLAGS_SPECIAL = $(CFLAGS)
+CFLAGS_SPECIAL += -ftrack-macro-expansion=0 -fno-diagnostics-show-caret
 ifeq ($(DEBUG), 1)
 CFLAGS += -g3 -gdwarf-2
 endif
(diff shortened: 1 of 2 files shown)


manthanand:
can you post a ss of the output when you make?

connorl309:
> can you post a ss of the output when you make?

image

Defines are specified as "DEFINES=<define name>=VALUE/<define name>=VALUE...."
Forward slash is the separator because Make functions cannot split on commas

manthanand:
I don't think this looks pretty. Also, there shouldn't be errors in the code?

manthanand:
Ig I wasn't too clear in the issue ticket. Could you make it so it prints out the values of all the #defines in the config file every time something is made?

connorl309:
image
There is no possible way in GCC to evaluate our ternary macros, so they will be printed as such.
Works with custom defines and still shows errors if you have invalid value combos such that compilation for other files dependent on the macros fails.

manthanand:
tis a shame. ok

manthanand:
I googled something, could you try this? `XSTR((int)NUM_MINIONS)` ?

connorl309:
> I googled something, could you try this? `XSTR((int)NUM_MINIONS)` for all the defines?

It just prefixes all the prints with "(int)"
image

manthanand:
Its mad stoopid if the ternaries are evaluated in runtime and not compile time.

BSP/STM32F413/Makefile
```diff
@@ -239,14 +250,13 @@ $(BUILD_DIR):
 .INTERMEDIATE: build_msg
 
 c_compile_msg:
-	@echo -e "${GREEN}Compiling C files...${NC}"
```
Champers5000, line 242: echo -e option is needed for escape characters to be interpreted
Why do yall keep getting rid of it ( ˘︹˘ )

Champers5000 (changes requested):

BSP/STM32F413/Makefile
connorl309: What escape characters do you need, the makefile works fine for me with -e being removed. It actually prints the -e on my machine.

connorl309: image

@Champers5000

Champers5000: image

That's what happens without the -e on my system
I'll see if there's a way to get it working on both machines

Champers5000: The escape characters are printed as is in text format and it looks wack af

Champers5000: @connorl309 What system are you compiling with

manthanand: this happens for me too which is why I get rid of it

manthanand: im guessing we are both doing it through wsl

Config/Src/config_msgs.c
```diff
@@ -16,56 +16,90 @@
...
 #define STR(x) #x
 
+#include "config.h"
+
+#ifdef NUM_MINIONS
+#pragma message ("NUM_MINIONS set to " XSTR(NUM_MINIONS))
+#else
+#pragma message ("NUM_MINIONS set to " XSTR(NUM_MINIONS))
+#endif
+
```
ray4477, line 26: why is `#pragma message ("NUM_MINIONS set to " XSTR(NUM_MINIONS))`

ray4477: set twice

BSP/STM32F413/Makefile
connorl309: Yes thru WSL

connorl309:
New changes pushed. should be finalized

BSP/STM32F413/Makefile
```diff
@@ -157,15 +157,21 @@ C_INCLUDES =  \
 # compile gcc flags
 ASFLAGS = $(MCU) $(AS_DEFS) $(AS_INCLUDES) $(OPT) -Wall -fdata-sections -ffunction-sections
 
-CFLAGS = $(MCU) $(C_DEFS) $(C_INCLUDES) $(OPT) -Wall -Werror -fdata-sections -ffunction-sections
+CFLAGS = $(MCU) $(C_DEFS) $(C_INCLUDES) $(OPT) -Wall -Werror -fdata-sections -ffunction-sections 
+CFLAGS_SPECIAL = $(CFLAGS)
+CFLAGS_SPECIAL += -ftrack-macro-expansion=0 -fno-diagnostics-show-caret
```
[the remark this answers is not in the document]
connorl309: let me try, probably but make variables are really finnicky

connorl309:
Ok so tests fail because some fixes from Champer's sim workflow branch aren't in here yet so some compilation explodes bc of array indices

manthanand (approved):

PR #593, opened 2023-11-11
add to readme for how to use DEFINES override

connorl309:
PR for quick readme update on how to use the DEFINES override for the makefile and BPS config


tiandahuang (commented):
i just found out that if you type `make help` it tells you how to do all this, if you add a mention to run make help then i can merge this real quick

tiandahuang (approved):

PR #601, opened 2024-03-21
Voltage skip assignment

Lakshay983:
Added an array in config.h that defines the distribution of cells monitored by each minion. Changed update_measurement to use this scheme.  Currently fixing open_wire to do the same.

My issue with how I currently do It as that I cannot check if the size of my distribution array = NUM_MINIONS defined. 

# Quality Assurance Checklist 

## Requirements 
- [ ] Followed Coding Style Guide 
- [X] Code Build checks pass 
- [X] No merge conflicts 
- [ ] If applicable, does software feature have simulation test associated with it (.json & out.json file)
- [X] Did you Have Fun?

## Things to Consider 


Config/Inc/config.h
```diff
@@ -101,8 +102,12 @@ typedef enum SafetyStatus_e {SAFE = 0, DANGER = 1, OVERVOLTAGE = 2, UNDERVOLTAGE
 #define MAX_VOLT_SENSORS_PER_MINION_BOARD	12	// User defined. The LTC6811 can actually measure 12 modules.
 #endif
 
+static const uint8_t VOLT_TAP_DIST[NUM_MINIONS] = {11, 10, 11}; // the distribution of modules each minion is monitoring
```
ray4477, line 105: Im not sure about this being in the config.h file. If it works it works but I'm pretty sure declarations of a static array should be done in voltage.c. not config.h as a lot of files pull from here. We can also just define # of taps per module in config.h

Apps/Src/Voltage.c
```diff
@@ -90,9 +91,21 @@ void Voltage_UpdateMeasurements(void){
...
     // package raw voltage values into single array
     for(uint8_t i = 0; i < NUM_BATTERY_MODULES; i++){
-        rawVoltages[i] = Minions[i / MAX_VOLT_SENSORS_PER_MINION_BOARD].cells.c_codes[i % MAX_VOLT_SENSORS_PER_MINION_BOARD];
+
+        // if we are out of modules in this minion to sample, move to the next minion
+        if(minionModulesRemaining <= 0) {
+            moduleIdx ++;
+            minionModulesRemaining = VOLT_TAP_DIST[moduleIdx];
+        }
+        rawVoltages[i] = Minions[moduleIdx].cells.c_codes[VOLT_TAP_DIST[moduleIdx] - minionModulesRemaining];
```
ray4477, line 107: This looks like it works, but honestly i cant see why we cant just do 3 `memcpy` into rawVoltages from the Minions.cells.c_codes array. It might be cleaner too.

BSP/STM32F413/Makefile
```diff
@@ -12,7 +14,11 @@ READABLE = true
 # optimization
 # change to -Og if you want the compiler to generate assembly that is very close
 # to your source code. Change to -O3 for best performance
+ifeq ($(debug), 0)
+OPT = -Og
```
ray4477, line 18: maybe include a comment on why we want to use -O3 for debugging

Apps/Src/Voltage.c
```diff
@@ -200,17 +219,26 @@ SafetyStatus Voltage_OpenWire(void){
     
     RTOS_BPS_MutexPend(&MinionsASIC_Mutex, OS_OPT_PEND_BLOCKING);
     
+    //int32_t openModules = LTC6811_run_openwire_multi(NUM_MINIONS, Minions, false);
     LTC6811_run_openwire_multi(NUM_MINIONS, Minions, false);
 
+    /*
+    To do: fix this to work with VOLT_TAP_DIST array
```
ray4477, line 226: bro has a todo in a pr. (i know it says draft pr)

Apps/Src/Voltage.c
tiandahuang: can't memcpy cuz datatype mismatch. inner for loop that essentially performs a memcpy works tho, would def be cleaner

Config/Inc/config.h
tiandahuang: better way of doing this (instead of static) is
```
__attribute__((weak)) const int 
VOLT_TAP_DIST[] = {11, 10, 11}; // the distribution of modules each minion is monitoring
_Static_assert((sizeof(VOLT_TAP_DIST)/sizeof(*VOLT_TAP_DIST)) == NUM_MINIONS, "error: ur cfg wrong");
```

Apps/Src/Voltage.c
ray4477: yeah u right

Apps/Src/Voltage.c
Lakshay983: Open wire is turned off and we don't need it for static monitoring, but there was no comment saying it was wrong.

BSP/STM32F413/Makefile
```diff
@@ -12,7 +14,11 @@ READABLE = true
 # optimization
 # change to -Og if you want the compiler to generate assembly that is very close
 # to your source code. Change to -O3 for best performance
+ifeq ($(debug), 0)
```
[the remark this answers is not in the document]
Lakshay983: u right, forgot the n in the ifneq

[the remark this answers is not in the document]
Lakshay983: do you have a suggestion for a better name? it's kinda bad since there's another variable named DEBUG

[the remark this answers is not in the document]
Lakshay983: just like "debugMode"

Config/Inc/config.h
IshDeshpa: agreed, this will check it before runtime right?

BSP/STM32F413/Makefile
```diff
@@ -3,7 +3,9 @@
 ######################################
 TARGET = bps-leader
 
-DEBUG = 0
+# sets debugging mode on or off based on BPS makefile
+DEBUG = $(debug)
```
IshDeshpa, line 7: Don't really understand the purpose of this. In the higher level makefile we export "debug", so it is available anywhere to use in this makefile. Why make another variable called "DEBUG" and then set it to "debug"?

imo you just call the one in the higher makefile DEBUG, and then remove this line. Since it's exported it should work the same.

BSP/STM32F413/Makefile
```diff
@@ -12,7 +14,11 @@ READABLE = true
 # optimization
 # change to -Og if you want the compiler to generate assembly that is very close
 # to your source code. Change to -O3 for best performance
+ifneq ($(debug), 0)
```
IshDeshpa, line 17: also maybe make this ifeq ($(debug) instead of ifneq ($(debug)) and reverse the order of the clauses. this doesn't actually matter it's just a personal preference ig.

BSP/STM32F413/Makefile
Lakshay983: u right, I'll change it.

Tasks/Src/Task_VoltTempMonitor.c
```diff
@@ -11,6 +11,12 @@
 //declared in Tasks.c
 extern cell_asic Minions[NUM_MINIONS];
 
+int piss = 0;
+#define PISS(...) \
+    if (piss % 32 == 0) { \
+        printf(__VA_ARGS__); \
+    }
+
```
Lakshay983, line 19: ah hell nah, u ain't gonna complain about my style guides then add a PISS macro

